PR #9, opened 2022-04-07
XRC-9: Composite content type

mkobetic:
Based on XIP-5 this XRC proposes a new content type `xtmp.org/composite:1.0` for multi-part messages where the parts can be of any content type and the composite parts can be arbitrarily nested.

Reference implementation https://github.com/xmtp/xmtp-js/pull/96

### DIFF
--- a/XIPs/xip-composite-content-type.md
+++ b/XIPs/xip-composite-content-type.md
@@ -14,7 +14,7 @@ Based on XIP-5 this XRC proposes a new content type `xtmp.org/composite:1.0` for
 
 ## Motivation
 
-Combining different kinds of content in a single messages seems obviously useful. Providing a generic facility to do that (similar to the MIME multipart content types) could satisfy some usecases, although more specific types designed to support particular use-cases may prove more popular in the long run. Especially since more specialized solutions will be easier to use (e.g. may infer content type rather then requiring content type IDs explicitly), and provide additional facitlities (e.g. provide hints about proper presentation of the content). Even if this generic type does not end up being widely used, it will also serve as an educational tool for future content type authors.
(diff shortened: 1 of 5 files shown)


neekolas (approved):
I really like how simple this is

XIPs/xip-composite-content-type.md
```diff
@@ -0,0 +1,83 @@
...
+}
+```
+
+## Rationale
+
+It is unclear at this point what particular aspects of multipart content will be relevant in practice, consequently this initial proposal is about as simple as it gets. Subsequent revisions can add more features, although there is some expectation that more specialized content types will be preferred for specific use-cases. Protobuf has been chosen to keep the implementation simple, which should support the educational goals of this proposal.
+
+## Reference Implementation
+
+The reference implementation is intended to be part of the XMTP SDK. [https://github.com/xmtp/xmtp-js/pull/96](https://github.com/xmtp/xmtp-js/pull/96). The codec will not be part of the default Client setup, clients wishing to employ this codec SHOULD register this codec explicitly.
```
saulmc, line 75: Given the educational goal of this XIP, would it make sense to provide an example of what it means to "register this codec explicitly"? Or is it obvious to the developer what to do?

jazzz (approved):
Extremely easy to follow. 👍

saulmc:
@mkobetic This should probably also include an update to the README that adds this XIP to the table.

saulmc (approved):

XIPs/xip-composite-content-type.md
mkobetic: Good point, I won't do it here since it applies to any codec, but I will document it in the SDK readme.

mkobetic:
> @mkobetic This should probably also include an update to the README that adds this XIP to the table.

@saulmc is it the author or the editor that should do that? With XIP-5 @mg0716 merged the PR and updated the table.

Speaking of which I need an editor for this XRC 🙏

mg0716:
> is it the author or the editor that should do that?

The editor (me) would be doing that.

mg0716 (approved):
nice

Looks good @mkobetic 

---

Additionally I updated the front matter to re-include a specific `xip` column, as unfortunately the `:` within the title column was preventing all front matter from rendering properly.

PR #14, opened 2023-01-31
XIP-14: Conversation context metadata schema

yash-luna:


XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,99 @@
...
+        brandInfo: {
+            displayName: 'Galaxy',
+            profileImage: 'galaxychat.com/assets/logo.png,
+            primaryColor: '#6865B8',
+        }
+    }
+}
+```
+
+The `profileImage` file is encouraged to meet the following criteria:
```
saulmc, line 54: Why SHOULD rather than MUST?

If this is a MUST, then we'd expect clients to validate and reject images that fail to meet the criteria. If a SHOULD, then they can decide at-will how to deal with non-conforming image.

XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,99 @@
+---
+xip: 10
```
saulmc, line 2: 14, since this is PR #14

XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,99 @@
...
+
+Example `brandInfo` implementation for a chat app named Galaxy
+
+```json
+{
+    conversationId: 'galaxy.chat/dm/uniqueIdentifier',
+    metadata: {
+        brandInfo: {
+            displayName: 'Galaxy',
+            profileImage: 'galaxychat.com/assets/logo.png,
```
saulmc, line 47: Are there any requirements or restrictions related to _where_ the image asset is hosted?

XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+}
+```
+
+The `profileImage` file is encouraged to meet the following criteria:
+
+- Aspect ratio: 1:1
+- Minimum resolution: 100x100
+- Maximum resolution: 800x800
```
neekolas, line 58: If these are externally hosted, why such a low maximum?

XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,99 @@
...
+}
+```
+
+New Lens schema
+
+```json
+{
+    conversationId: 'lens.dev/dm/${memberA}-${memberB}',
+    metadata: {
+        brandInfo: {
```
neekolas, line 84: Metadata is currently restricted as a mapping that only accepts strings as values. We cannot have nested objects.

XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,99 @@
...
+            profileImage: 'lens.xyz/assets/myimage.png,
+            primaryColor: '#ffffff',
+        }
+    }
+}
+```
+
+## Security Considerations
+
+The metadata field can be spoofed by malicious apps to display Names and Images that degrade the user experience and are harmful to the brand’s perception. A mechanism for apps to sign messages and related metadata can enable front-ends to verify if the signature is from a credible source in order to prevent such spoofing. An affordance for client verifiability is under consideration.
```
mkobetic, line 95: The invitation is encrypted and authenticated using the sender & recipient keys. So the metadata must have been created by one of the parties if the invitation is successfully unsealed (https://github.com/xmtp/proto/blob/main/proto/message_contents/invitation.proto).

XIPs/xip-10-brandinfo-schema.md
yash-luna: It MUST!

XIPs/xip-10-brandinfo-schema.md
yash-luna: No. Should there be?

XIPs/xip-10-brandinfo-schema.md
yash-luna: Based on Darick's suggestions. Thoughts on increasing min resolution @darickdang?

XIPs/xip-10-brandinfo-schema.md
yash-luna: - What was the rationale behind allowing strings only? 
- Can we relax restrictions and allow nested objects? If not, what's the best way to format the brandInfo as a string?

XIPs/xip-10-brandinfo-schema.md
mkobetic: I see, I misread this section. This is however a more general concern that currently apps have complete access to a user's XMTP identity/keys, so a users must necessarily trust the app they are using in all regards not just this particular aspect. It may still be worth mentioning here, but I'm less sure about what the solution should be. Some amount of trust between the user and the app may always be necessary.

XIPs/xip-10-brandinfo-schema.md
neekolas: > What was the rationale behind allowing strings only?

This is data that is going to be interpreted by many different apps who aren't going to be able to share a schema. If we left it free form, it opens up new avenues for client-side bugs (you start sending `brandinfo` as an object, some other app starts sending `brandinfo` as an array, and it's easy for one of your apps to break when it encounters a conversation from the other). I would think of this as something closer to HTTP headers than a free-form datastore. 

> Can we relax restrictions and allow nested objects? If not, what's the best way to format the brandInfo as a string?

It's baked into the proto schema, so it would be a significant change if we thought it was advisable (which I still don't).

One way to namespace the brandinfo would be to use dots in the keys. For example `brandinfo.displayName`. Could also embed a JSON object as `brandinfo`, but I'm less partial to that.

XIPs/xip-10-brandinfo-schema.md
darickdang: @neekolas Since the image would be used in avatar-like settings (or to complement an avatar), I looked at popular services/libraries such as Gravatar (minimum size 1px) and Libravatar (minimum size is 80x80) standards in addition to Apple's Human Interface Guidelines of having a minimum toucharea of 44x44 (and then I doubled that to add @2x support as the minimum) to create the number you see here.

Open to making it bigger though. Not that strong of an opinion from my end here.

neekolas: I just wonder if some apps are going to want a clickable/zoomed view of app icons. Or an extra large icon. Maybe that's not a real concern here.

XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,81 @@
...
+created: 2023-01-31
+---
+
+## Abstract
+
+This XIP proposes a schema for sharing brand information as it relates to `conversationId`. A standard way to represent this information in conversation `metadata` will enable different front-ends to surface the context captured by `conversationId` in a consistent, user-friendly manner while still affording clients design and UI flexibility.
+
+## Motivation
+
+The SDK API currently accepts a unique `conversationId` per address pair and allows `metadata` to be set as well. Today, different apps use these parameters in different ways and therefore other apps may not know how to render the `conversationId` appropriately within the UI to reflect the corresponding brand. We propose a schema to standardize the use of conversation `metadata` for the purpose of client-side brand expression stemming from the `conversationId`. The `metadata` field may include other non-standard (or future standard) properties.
```
jhaaaa, line 18:
```suggestion
The SDK API currently accepts a unique `conversationId` per address pair and allows `metadata` to be set as well. Today, different apps use these parameters in different ways. Therefore an app may not know how to correctly render in its UI another app's brand expressed in its `conversationId` and `metadata`. We propose a schema to standardize the use of conversation `metadata` for the purpose of app brand expression stemming from the `conversationId`. The `metadata` field may include other non-standard (or future standard) properties.
```

Not sure if this helped - but trying to clearly express one app not knowing how to interpret the conversationId and metadata from another app.

XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,81 @@
...
+
+Note that the schema is only applicable when setting a non-null `conversationId`. If `conversationId` is null, the SDK does not allow setting `metadata`.
```
jhaaaa, line 20: A great detail - I need to add this to the docs!

XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,81 @@
...
+## Specification
+
+Proposed `brandInfo` schema in conversation `metadata`
+
+```json
+{
+    conversationId: "mydomain.xyz/abc/qrs",
+    metadata: { "brandInfo.displayName": "My company", "brandInfo.profileImage": "mydomain.xyz/assets/myimage.png", "brandInfo.primaryColor": "#ffffff" }
+}
+```
```
jhaaaa, line 31: What do you think of formatting it like this to make the metadata schema a little bit more readable? It's so good I want people to immediately see the value!

```
{
   "conversationId":"mydomain.xyz/abc/qrs",
   "metadata":{
      "brandInfo.displayName":"My company",
      "brandInfo.profileImage":"mydomain.xyz/assets/myimage.png",
      "brandInfo.primaryColor":"#ffffff"
   }
}
```

XIPs/xip-10-brandinfo-schema.md
jhaaaa: I wonder if it might help to include a mockup of a UI showing how the `displayName`, `profileImage`, and `primaryColor` might be used? While the `profileImage` might be more obvious, people might be curious about where the `displayName` and `primaryColor` might be used? An image might also really drive home the value! 

And is the idea now that we don't need to use only the `conversationId` as conversation label text?

XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,81 @@
...
+The `profileImage` must meet the following criteria:
+
+- Aspect ratio: 1:1
+- Minimum resolution: 100x100
+- Maximum resolution: 800x800
+- Format: .PNG, .WEBP
+
+The `primaryColor` must be a hex color code.
+
+## Backwards Compatibility
```
jhaaaa, line 51: SUPER nit.......

```suggestion
## Backward compatibility
```

XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,81 @@
...
+
+Existing conversations will not be affected by the adoption of the new schema; only conversations created after the new schema is finalized will follow the updated schema.
+
+To maintain backwards compatibility, clients may consider rendering popular but older schemas (such as Lens), present in conversations created before the finalization of this XRC, in a manner consistent with their updated schema that follows this XRC.
```
jhaaaa, line 55: Super nit again....... =)

Wondering - "clients may..." or "apps may..."?

```suggestion
To maintain backward compatibility, clients may consider rendering popular but older schemas (such as Lens), present in conversations created before the finalization of this XRC, in a manner consistent with their updated schema that follows this XRC.
```

XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,81 @@
...
+New Lens schema
+
+```json
+{
+    conversationId: "lens.dev/dm/${memberA}-${memberB}"
+    metadata: { "brandInfo.displayName": "Lens", "brandInfo.profileImage": "lens.xyz/assets/myimage.png", "brandInfo.primaryColor": "#ffffff" }
+}
+```
+
+## Security Considerations
```
jhaaaa, line 75: Nit again...

```suggestion
## Security considerations
```

XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,81 @@
...
+
+The metadata field can be spoofed by malicious apps to display Names and Images that degrade the user experience and are harmful to the brand’s perception. A mechanism for apps to sign messages and related metadata can enable front-ends to verify if the signature is from a credible source in order to prevent such spoofing. An affordance for client verifiability is under consideration.
```
jhaaaa, line 77:
```suggestion
The metadata fields can be spoofed by malicious apps to display names and profile images that degrade the user experience and can be harmful to a brand’s perception. A mechanism for apps to sign messages and related metadata can enable frontends to verify if the signature is from a credible source to prevent such spoofing. An affordance for client verifiability is under consideration.
```

jhaaaa (commented):
So excited about this XRC, @yash-luna - thank you for designing and articulating it so clearly!  Added a few comments for your consideration -- please use only what you find helpful!

XIPs/xip-10-brandinfo-schema.md
yash-luna: Adding a sample UI is such a good idea @jhaaaa! 

Yes, with a conversationId like 'lens.dev/dm/...' there isn't a shared understanding of how to render it? Should the UI display 'Lens', 'lens.dev', or the complete conversationId? The displayName makes it explicit

XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,81 @@
...
+author: Yash Lunagaria (@yash-luna)
+status: Review
+type: Standards Track
+category: XRC
+created: 2023-01-31
+---
+
+## Abstract
+
+This XIP proposes a schema for sharing brand information as it relates to `conversationId`. A standard way to represent this information in conversation `metadata` will enable different front-ends to surface the context captured by `conversationId` in a consistent, user-friendly manner while still affording clients design and UI flexibility.
```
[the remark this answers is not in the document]
yash-luna: I like the suggestion of making the guidance obvious but branding here isn't necessarily 'app branding'. Taking Lens as an example, the branding for Lens conversationIds maps to the Lens protocol, not a specific Lens app. Of course, these properties can be used by an app like Converse that creates their own conversationIds. They can also be used to create a conversationId about a specific NFT offer/sale/bid/transaction between 2 parties. 

What can be better names for these properties? I'd appreciate some @jhaaaa magic here 😄 

Perhaps displayInfo (vs brandInfo), prettyName (vs displayName), profileImage, and primaryColor?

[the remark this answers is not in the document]
yash-luna: XRC is a type of XIP. I'm also trying to wrap my head around these terms. 

@neekolas pls correct me if I'm off. My understanding is:

`conversationId` and `metadata` are properties of a conversation/topic. There is active debate around whether a conversation can have multiple topics. We're leaning towards not allowing that.

XIPs/xip-10-brandinfo-schema.md
darickdang: Great call @jhaaaa. Will look at making an example.

XIPs/xip-10-brandinfo-schema.md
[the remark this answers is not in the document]
jhaaaa: Thank you for sharing this information @yash-luna - so helpful. As we discussed, sharing this proposed rev to the opening paragraph that removes the need to say "brand."

>This XIP proposes a schema for sharing additional context about a conversation beyond a `conversationId`. Having a standard way to provide not only a `conversationId`, but also conversation `metadata` will enable different frontends to display useful contextual information in a consistent, user-friendly manner, while still affording apps design and UI flexibility.

XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,90 @@
...
+## Specification
+
+Proposed `displayInfo` schema in conversation `metadata`
+
+```json
+{
+    conversationId: "mydomain.xyz/abc/qrs",
+    metadata: { 
+        "displayInfo.prettyName": "My company", 
+        "displayInfo.profileImage": "mydomain.xyz/assets/myimage.png",
```
jazzz, line 31: How does `displayInfo.profilePicture` differ from a `Contact ProfilePicture` or a `group conversation image`? Is the intention different?

XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+Example `displayInfo` implementation for a chat app named Galaxy
+
+```json
+{
+    conversationId: "galaxy.chat/dm/uniqueIdentifier",
+    metadata: { 
+        "displayInfo.prettyName": "Galaxy", 
+        "displayInfo.profileImage": "galaxychat.xyz/brandassets/logo.png",
```
jazzz, line 43: What is the rationale behind embedding this data into the Conversation Invites? 

Why does the invite not contain a pointer to a central registry, which would be easier for updates?
Why is this information not contained in a metamessage within the conversation?

jazzz (commented):
I'm not sure the Motivation is clear to me. 
The examples and text hint towards standardizing Client applications metadata{AppName + Photo} . However the schema naming seems to hint towards conversation metadata {GroupChat Title, GroupChat ProfileImage}

XIPs/xip-10-brandinfo-schema.md
yash-luna: This XIP started with the idea of brandInfo/clientMetadata in mind but we realized that the application is broader - i.e. to add context beyond conversationId that allows a frontend to contextualize conversations

Sample UIs and non-clientMetadata examples are work in progress. Sharing most recent UI mock that highlights how they can be utilized for clientMetadata, NftMetadata, or other metadata that contextualizes the conversation.

image

yash-luna:
> I'm not sure the Motivation is clear to me.

Sharing UI examples to provide clarity. 

Overall, you comment inspired me to think whether we should aim for a generic conversation metadata schema or have different schemas catering to different use-cases (clientMetadata, NftMetadata, groupchatMetadata, etc)?

XIPs/xip-10-brandinfo-schema.md
jazzz: This example is very helpful. 

I might suggest that we scope the properties to something more descriptive than `DisplayInfo.DisplayName`. In particular we should create specific terms around what to collectively call the OpenSea/Lens tags.

e.g. `ConversationSource.DisplayName`, `ConversationOrigin.DisplayName`, `AppSource....`

My main thought here is that with groupchats coming in the near future, `DisplayName` will become a loaded term.

XIPs/xip-10-brandinfo-schema.md
jazzz: I think these are valid reasons to embedded into the conversation metainfo.  I can get behind this. 

In my mind a central registry could be ENS with values stored as textfields. It would address Longterm availability (at least in the context of apps), and Ease of use (Passing update costs to the brands/applications). I think consistency will be an issue regardless and a well defined spec is needed.

XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,90 @@
...
+Note that the schema is only applicable when setting a non-null `conversationId`. If `conversationId` is null, the SDK does not allow setting `metadata`.
+
+## Specification
+
+Proposed `displayInfo` schema in conversation `metadata`
+
+```json
+{
+    conversationId: "mydomain.xyz/abc/qrs",
+    metadata: {
```
jazzz, line 29: We may want to consider adding an explicit version to this schema, so future updates are easier to deploy.

The scheme could be directly tied to conversationVersion, But then schema changes would result in a bump in both. 

Perhaps something explicit like:
```
metadata: {
   "conversationSource_V1": {
       "prettyName": "My company",
       "profileImage": "mydomain.xyz/assets/myimage.png"
   }
}
```

That would allow the SDK to add multiple Objects to the metadata if older clients need to be supported

XIPs/xip-10-brandinfo-schema.md
galligan, line 31: This should also include an affordance for an image URI. For web3-native projects they may have an IPFS, IPLD or other resource URI as a default, with a specific image URL as secondary.

This would provide a front-end with the ability to use a URI directly.

XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,90 @@
...
+    metadata: { 
+        "displayInfo.prettyName": "Lens", 
+        "displayInfo.profileImage": "lens.xyz/assets/myimage.png", 
+        "displayInfo.primaryColor": "#ffffff" }
+}
+```
+
+## Security considerations
+
+The `metadata` fields can be spoofed by malicious apps to display names and profile images that degrade the user experience and harm brand perception. A mechanism for apps to sign payloads such as conversation metadata and messages can enable frontends to verify the sending client's identity and prevent such spoofing. An affordance for client verifiability is under consideration.
```
galligan, line 86:
```suggestion
A brand's chosen metadata fields and content may be co-opted by other apps to display the same names and profile images which could impact the user experience and perception of a message's origin or affiliation. A mechanism for apps to sign payloads such as conversation metadata and messages could enable frontends to verify the sending client's metadata authenticity and prevent such spoofing. An affordance for verifying metadata could be considered in a future proposal.
```

I'm suggesting that we lighten the language on "spoofing" and "malicious" as it's simply an open space, and there's nothing currently preventing a project from co-opting some labels. Not ideal, but we should be careful about using such pointed language as it may diminish its value in other circumstances.

galligan (commented):
This is a great start, @yash-luna. Would love to have some of the thoughts in this review to be considered. Happy to have a conversation about some of them and go deeper.

galligan:
> This is somewhat tangential, but have we ever considered changing the term `conversationID`? I was confused by this when I first started because it sounds like a globally unique identifier for an individual conversation, rather than an identifier for a set of conversations. Better terms IMO would be something like `conversationType` or `appID`

@richardhuaaa It may be useful to include those as additional fields in some fashion, perhaps as the metadata. For example, an app may choose to have multiple conversations between the same parties present. Or perhaps it might attempt to use them to separate "threads". This may not be the intended use but for now that's how it could happen.

XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,90 @@
...
+{
+    conversationId: "galaxy.chat/dm/uniqueIdentifier",
+    metadata: { 
+        "displayInfo.prettyName": "Galaxy", 
+        "displayInfo.profileImage": "galaxychat.xyz/brandassets/logo.png", 
+        "displayInfo.primaryColor": "#6865B8" }
+}
+```
+
+The `profileImage` must meet the following criteria:
```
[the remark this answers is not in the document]
yash-luna: It originally said 'should' and was changed to 'must' based on @saulmc's comment. I don't have a strong opinion and am ok with changing the verbiage.

XIPs/xip-10-brandinfo-schema.md
```diff
@@ -0,0 +1,90 @@
...
+
+- Aspect ratio: 1:1
+- Minimum resolution: 100x100
+- Maximum resolution: 800x800
+- Format: .PNG, .WEBP
+
+The `primaryColor` must be a hex color code.
```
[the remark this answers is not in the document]
yash-luna: How is a fallback color different from a second hex code? Didn't understand how 'a fallback color would allow the front-end to set up their own interpretation of the colors'

XIPs/xip-10-brandinfo-schema.md
yash-luna: Can the URI be set as a value to profileImage?

petermdenton: > This should also include an affordance for an image URI. For web3-native projects they may have an IPFS, IPLD or other resource URI as a default, with a specific image URL as secondary.
> 
> This would provide a front-end with the ability to use a URI directly.

If all of these images are externally hosted, isn't that going to add several additional round trip requests for the browser?

XIPs/xip-10-brandinfo-schema.md
galligan: @richardhuaaa agree on the registry component for sure.

I do wonder if somehow the metadata itself can be hashed, and the hash compared against the registry, to know what's authentic. And perhaps the conversationId could be used as a checksum against the metadata, like how you might with a barcode.

XIPs/xip-10-brandinfo-schema.md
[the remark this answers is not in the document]
galligan: Must to me implies that somehow it is enforceable. I'm not sure I see how that would be possible. Should feels more appropriate in this situation but happy to discuss further with @saulmc.

XIPs/xip-10-brandinfo-schema.md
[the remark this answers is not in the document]
galligan: @yash-luna to use an analogy, a "fallback" color in this case is the description of a color of paint, whereas the hex code would be similar to the paint code instructions on the outside of a can of paint.

In the fallback case the specific color used would be based off of the developer's interpretation. With the hex color there is no interpretation and it is explicit.

The naming scheme for the fallback could be based off of an existing example, such as Material Design or Tailwind CSS's color palette. It usually has a color name and separate value to indicate darkness or saturation.

If given the choice, if I were developing an app I would always pick the fallback color as I'd have a bit more control over how it should be represented. Without that control, I might ignore the color altogether as the results could really clash with my experience.

XIPs/xip-10-brandinfo-schema.md
jazzz: @galligan  I like the explicit structure of version, type, content.

I might suggest a structure that looks like:

```
metadata: [
	{
		"type": "conversation_source",
		"version": "1.0" 
		"content": {
		  	"title": "My company",
    		"description"; "My description that may be used in an app",
    		"image": "https://mydomain.xyz/assets/myimage.png",
    		"imageURI": "ipfs://zafybeict2kq6gt4ikgulypt7h7nwj4hmfi2kevrqvnx2osibfulyy5x3hu/myimage.png"
		}
	},
	{
		"type": "group_info",
		"version": "1.0" 
		"content": {
		  	"name": "Frens",
    		"admin"; "....",
    		"image": "ipfs://pkrewapct2kq6gt4ikgulypt7h7nwj4hmfi2kevrqvnx2osibfulypkl39a/myimage.png".
		}
	}
]
```

- There should be room for multiple metadata objects
- Objects names ought to be hyper specific to the data it is representing. `Conversation.Image` is ambiguous. Codifying clear names for entities will help developers in the future.

XIPs/xip-10-brandinfo-schema.md
galligan: > Can the URI be set as a value to profileImage?

@yash-luna it could be, but setting URI as a separate value to URL would have its benefits. For clients that don't support the protocol the URI is pointing to, then it can look at the URL instead.

> If all of these images are externally hosted, isn't that going to add several additional round trip requests for the browser?

@petermdenton Possibly. But it can be accounted for. And otherwise I'm not sure where these images will be coming from except through external hosting. Encouraging them to be stored in-protocol may be a slippery slope that we don't want to go down.

With including the URI and using IPFS, a client could easily cache images and maintain integrity without trouble. An image would be fetched via its URI, and could be saved locally with its CID intact such that the image could not be altered without also changing its CID.

XIPs/xip-10-brandinfo-schema.md
neekolas: Worth noting that neither an array or nested object are possible with today's metadata. It is strictly a Map<string, string>. Closer to HTTP headers than a free form JavaScript object.

yash-luna: Based on the suggestions above, there's benefits to the protocol and DX to enable metadata to take arrays and/or nested objects. It would enable versioning for metadata schemas and make it friendly for other use cases beyond conversationId context.

@neekolas @jazzz: What are the downsides of such a change? Can it be in scope for xmtp-js v8?

jhaaaa:
No activity for over 6 months. Set status to Stagnant and merging.

PR #15, opened 2023-02-16
Attachment content type proposal

nakajima:
Here's a proposal for adding a very basic content type for attachment messages on XMTP.

### DIFF
--- a/XIPs/zip-15-attachment-content-type.md
+++ b/XIPs/zip-15-attachment-content-type.md
@@ -35,12 +35,11 @@ The message MUST include the following parameters:
 ```ts
 {
   mimeType: string,
-  filename: string
+  filename: string,
 }
 ```
 
-
-The content of the encoded message is arbitrary data. It's up to clients to use the mime type and filename to determine how to render the message.
+The content of the encoded message is arbitrary data. It's up to clients to use the mime type and filename to determine how to render the message. The content MUST not be more than 1MB.
 
 ## Backward compatibility
 
@@ -48,8 +47,8 @@ Clients encountering messages of this type must already be able to deal with mes
 
 ## Reference implementation
 
-Implementation reference: https://github.com/xmtp/xmtp-ios/pull/62
-Client Usage reference: https://github.com/xmtp-labs/xmtp-inbox-ios/pull/65
+- [Implementation reference](https://github.com/xmtp/xmtp-ios/pull/62)
+- [Client Usage reference](https://github.com/xmtp-labs/xmtp-inbox-ios/pull/65)
 
 ## Security considerations
 
--- a/XIPs/zip-15-attachment-content-type.md
+++ b/XIPs/zip-15-attachment-content-type.md
@@ -1,0 +1,60 @@
+---
+xip: 15
+title: Attachment Content Type
+description: Content type for basic attachments
+author: Pat Nakajima
+status: Review
+type: Standards Track
+category: XRC
+created: 2023-02-15
+---
+
+## Abstract
+
+This XRC proposes a new content type for messages that do not consist of text. It attempts to include only the bare minimum for clients to determine how to display such messages in order to provide flexibility for more rich types in the future.
+
+## Motivation
+
+Not everything in a conversation is text. MMS, email attachments, gifs, etc are all examples of times when communications have gone beyond words. In order to provide users a real messaging platform, XMTP needs to provide this capability.
+
+## Specification
+
+Proposed content type:
+
+```js
+{
+  authorityId: "xmtp.org"
+  typeId: "attachment"
+  versionMajor: 0
+  versionMinor: 1
+}
+```
+
+The message MUST include the following parameters:
+
+```ts
+{
+  mimeType: string,
+  filename: string
+}
+```
+
+
+The content of the encoded message is arbitrary data. It's up to clients to use the mime type and filename to determine how to render the message.
+
+## Backward compatibility
+
+Clients encountering messages of this type must already be able to deal with messages of an unknown content type, so whatever considerations they're making there should work here too.
+
+## Reference implementation
+
+Implementation reference: https://github.com/xmtp/xmtp-ios/pull/62
+Client Usage reference: https://github.com/xmtp-labs/xmtp-inbox-ios/pull/65
+
+## Security considerations
+
+Having arbitrary data anywhere can be risky, but this is already the case for our messages, since there's no server side validation of message contents (besides size). The same protections we have now would be in place while the same pitfalls we have would still be there as well.
+
+## Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).


XIPs/zip-15-attachment-content-type.md
```diff
@@ -0,0 +1,60 @@
...
+created: 2023-02-15
+---
+
+## Abstract
+
+This XRC proposes a new content type for messages that do not consist of text. It attempts to include only the bare minimum for clients to determine how to display such messages in order to provide flexibility for more rich types in the future.
+
+## Motivation
+
+Not everything in a conversation is text. MMS, email attachments, gifs, etc are all examples of times when communications have gone beyond words. In order to provide users a real messaging platform, XMTP needs to provide this capability.
```
neekolas, line 18: I can get my head around a generic "attachment" type, if we are to treat these like files.

I'm less confident we should be relying on MIME content type alone for content meant to be rendered inside client apps. It feels like without Content-Disposition and Content-Transfer-Encoding we are going to run into the same problems that necessitated those being added to MIME.

For media types being rendered inside an application there are likely more fields than just `mimeType` and `filename` that we would want. For example, an image type might want `size`, `alt-text`, `attribution`, etc. Cases like those might be better served with bespoke XMTP `ContentType`s

XIPs/zip-15-attachment-content-type.md
```diff
@@ -0,0 +1,60 @@
...
+author: Pat Nakajima
+status: Review
+type: Standards Track
+category: XRC
+created: 2023-02-15
+---
+
+## Abstract
+
+This XRC proposes a new content type for messages that do not consist of text. It attempts to include only the bare minimum for clients to determine how to display such messages in order to provide flexibility for more rich types in the future.
```
neekolas, line 14: Given the results of [our most recent benchmarks](https://github.com/xmtp/xmtp-js/pull/273), I wonder how hard we should be pushing client developers to attach large-ish attachments to messages.

XIPs/zip-15-attachment-content-type.md
nakajima: > It feels like without Content-Disposition and Content-Transfer-Encoding we are going to run into the same problems that necessitated those being added to MIME.

I think encoding could definitely be worth adding. I'm not as clear where disposition would be as necessary here but could just be misunderstanding.

> For example, an image type might want size, alt-text, attribution, etc. Cases like those might be better served with bespoke XMTP ContentTypes

Totally agree that we could have some much richer types. I think this type would be a decent bare bones solution that gets us unblocked on a lot of functionality for now, if we ever got to a place where we had a more complete set of bespoke types, we could look at deprecating this. I think that's a much longer process though.

XIPs/zip-15-attachment-content-type.md
neekolas: If the client wants to store it on our network, it has to be a message. I could imagine a world where we generate one-off encryption keys and store those in a message, and then use those keys to decrypt content stored elsewhere. That would be E2EE but would allow attachments of unlimited size.

Of course, then we have to figure out where and how to host the attachment.

XIPs/zip-15-attachment-content-type.md
yash-luna: Can fields like size, alt-text, etc be added as optional parameters to support richer types from the get go?

XIPs/zip-15-attachment-content-type.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+The message MUST include the following parameters:
+
+```ts
+{
+  mimeType: string,
+  filename: string
+}
+```
```
yash-luna, line 40: Some noob (I don't have a lot of MIME context) questions that I'm thinking about:

- What values are expected for mimeType and filename? Is the list of values for mimeType discrete or can it be anything? Should we reference something like the Mozilla list [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types).
- Should these strings be formatted in a particular way? If yes, how?
- What is the filename used for?
- How should frontend clients use the mimeType and filename info when rendering? The reference implementation is helpful but I'm wondering where frontends should follow a certain standard vs have the option to be opinionated?

XIPs/zip-15-attachment-content-type.md
yash-luna: In a future with bespoke content types for Image, Audio, Video, etc when should apps use an attachment content type vs a dedicated content type (e.g. image)?

Trying to understand if attachment type has a use case in that future. Or if it's certain that it'll be deprecated.

XIPs/zip-15-attachment-content-type.md
nakajima: I think that could be a rad new content type as well, something like a remote attachment. I still think there's probably room for smaller attachments that are inline. Sort of like how iMessage lets you send pics and videos up to a certain size, then suggests uploading to iCloud for larger stuff.

XIPs/zip-15-attachment-content-type.md
nakajima: > What values are expected for mimeType and filename? Is the list of values for mimeType discrete or can it be anything? 

[MIME types](https://en.wikipedia.org/wiki/Media_type) are a pretty good standard for describing content types. It's possible someone could just put garbage in that field and clients wouldn't know what to do, but that's a possibility with any content description scheme. Clients should be able to handle corrupt messages for any number of reasons.

> Should these strings be formatted in a particular way? If yes, how?

It's just the type and subtype separated by a `/`. We can add some more examples here.

> What is the filename used for?

I added it because it's a good hint for what to name the file if a client wants to save it. It could also be used to display inline with the message.

> How should frontend clients use the mimeType and filename info when rendering?

To start out, they can look to see if the mime type is an image and show an image, if there's a type the client doesn't know how to render, they can do [a generic thing like iMessage](https://user-images.githubusercontent.com/483/219428312-1d132d74-4d63-4b9b-b1a7-c155c2e553f0.png).

> The reference implementation is helpful but I'm wondering where frontends should follow a certain standard vs have the option to be opinionated?

I think this is a case where clients can be opinionated. As we introduce more rich content types, we can try to give more guidance on usage.

XIPs/zip-15-attachment-content-type.md
neekolas: I agree with @nakajima that we don't want to throw in every possible configuration to this generic content type. If we want an image type with alt-text, we should build an image content type that has the appropriate fields. That way a client can support the image content type fully or not at all.

XIPs/zip-15-attachment-content-type.md
nakajima: Wrote a proposal for a remote attachment type here: https://github.com/xmtp/XIPs/pull/18.

nmalzieu: Just wondering, if I send a 1MB image on the network, it will be part of the encoded `message` field on the `envelope.message` field right?
Since we send the envelope via notification to the mobile apps, should we take into account the size limit of notification payloads?

PR #16, opened 2023-02-17
Image/Video url content type

jsonpreet:
Hi Everyone, Here's my proposal for Add Image Url Content Type & Video Url Content Type in XMTP.

### DIFF
--- a/XIPs/xip-16-image-url-content-type.md
+++ b/XIPs/xip-16-image-url-content-type.md
@@ -22,11 +22,40 @@ As per users requirements, and for my platform I decided to create a new content
 Proposed content type:
 
 ```js
+new ContentTypeId({
+    authorityId: "xmtp.org",
+    typeId: "media-image-url",
+    versionMajor: 1,
+    versionMinor: 0,
+})
+
+class ImageCodec {
+    get contentType() {
+        return ContentTypeImageKey
+    }
+
+    encode(content: string): EncodedContent {
+        return {
+            type: ContentTypeImageKey,
+            parameters: {},
+            content: new TextEncoder().encode(content),
+        }
+    }
+
+    decode(content: EncodedContent): string {
+        // console.log(content.content.toString())
+        const uint8Array = content.content
+        const key = new TextDecoder().decode(uint8Array)
+        return key
+    }
+}
+```
+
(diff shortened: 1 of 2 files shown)


XIPs/xip-16-image-url-content-type.md
```diff
@@ -0,0 +1,54 @@
...
+{
+  authorityId: "xmtp.org"
+  typeId: "media-image-url"
+  versionMajor: 0
+  versionMinor: 1
+}
+```
+
+
+The content of the encoded message is arbitrary data. The content MUST not be more than 1MB.
```
neekolas, line 34: Thinking about accessibility, we probably want to support an optional `alt-text` field for screen readers.

neekolas, line 34: I would think if the content is only a URL, we could safely restrict it to something much smaller than 1MB. Even 10kb would be very generous for most URLs, and longer than what some browsers support. Edge for example limits to ~2k characters.

We can also specify that the URL must be encoded as UTF-8 text or the content would be rejected, which would remove some potential for misuse.

yash-luna, line 34: @jsonpreet how can we support end-to-end encrypted image sharing that sends the data over the XMTP network instead of just send the URL?

neekolas: That feels like it should be a different content type altogether. I can see use-cases for both.

jsonpreet: Agreed on URL size limit. 

On Encoding I am using standard TextEncoder
```
class ImageCodec {
    get contentType() {
        return ContentTypeImageKey
    }

    encode(content: string): EncodedContent {
        return {
            type: ContentTypeImageKey,
            parameters: {},
            content: new TextEncoder().encode(content),
        }
    }

    decode(content: EncodedContent): string {
        // console.log(content.content.toString())
        const uint8Array = content.content
        const key = new TextDecoder().decode(uint8Array)
        return key
    }
}
```

jsonpreet: @yash-luna For end-to-end image encryption we can use base64 image data. This content type is build for send encoded image URL over the XMTP Network.

PR #17, opened 2023-02-17
Clarify that xmtp.org is a reserved authorityId value

jhaaaa:
Based on developers building custom content types that use `xmtp.org` as the `authorityId`, we thought it might be good to update XIP-5's definition of the authority ID to state that the `xmtp.org` `authorityId` value is reserved for standard content types.


mkobetic (approved):
Love it. This was definitely an unfortunate omission.

PR #18, opened 2023-02-20
Remote Attachment Content Type Proposal

nakajima:
A suggestion for how we can handle larger messages but maintain E2E encryption.

### DIFF
--- a/XIPs/xip-17-remote-attachment-content-type-proposal.md
+++ b/XIPs/xip-17-remote-attachment-content-type-proposal.md
@@ -44,11 +44,14 @@ The encoded content MUST have the following parameters:
 	salt: string,
 	
 	// A hex string for the nonce used to encrypt the remote content payload
-	nonce: string
+	nonce: string,
+	
+	// The scheme of the URL. Must be "https://"
+	scheme: "https://"
 }
 ```
 
-The content of the encoded message is a URL that points to an encrypted `EncodedContent` object. The `EncodedContent`'s content type MUST not be another `RemoteAttachment`.
+The content of the encoded message is a URL that points to an encrypted `EncodedContent` object. The content MUST be accessed by an HTTP `GET` request to the URL. The `EncodedContent`'s content type MUST not be another `RemoteAttachment`.
 
 By using `EncodedMessage`, we can make it easier for clients to support any message content already used on the network (with the exception of `RemoteAttachment` as mentioned above).
 
--- a/XIPs/xip-17-remote-attachment-content-type-proposal.md
+++ b/XIPs/xip-17-remote-attachment-content-type-proposal.md
@@ -41,7 +41,7 @@ Clients encountering messages of this type must already be able to deal with mes
 ## Reference implementation
 
 - [Implementation reference](https://github.com/xmtp/xmtp-ios/pull/68)
-- Client Usage reference (TODO if this is a direction we're interested in pursuing)
+- [Client Usage reference](https://github.com/xmtp-labs/xmtp-inbox-ios/pull/83)
 
 ## Security considerations
 
--- a/XIPs/xip-17-remote-attachment-content-type-proposal.md
+++ b/XIPs/xip-17-remote-attachment-content-type-proposal.md
@@ -11,7 +11,7 @@ created: 2023-02-15
 
 ## Abstract
 
-This XRC proposes a new content type for messages that do not consist of text and/or are too large to be sent on the XMTP network. It builds on [XIP 15](https://github.com/xmtp/XIPs/pull/15)'s basic attachment type, but instead of messages being sent on the network, the message contains a URL. The contents of the URL is an `Envelope` that can be decoded by clients.
+This XRC proposes a new content type for messages that do not consist of text and/or are too large to be sent on the XMTP network. It builds on [XIP 15](https://github.com/xmtp/XIPs/pull/15)'s basic attachment type, but instead of messages being sent on the network, the message contains a URL. The contents of the URL is an encrypted protobuf encoded `EncodedContent` that can be decrypted and decoded by clients.
 
 ## Motivation
 
@@ -30,9 +30,29 @@ Proposed content type:
 }
 ```
 
-The content of the encoded message is a URL that points to encoded Envelope data that would not be limited by the 1MB limit, since it's not being sent on the network. By using the existing `Envelope` format, we can still ensure E2E encryption for this data.
+The encoded content MUST have the following parameters:
 
-The envelope itself can contain any type of message already allowed on the network. The reference implementation uses the `Attachment` type from XIP 15, but if we introduce richer types for things like images or video, those would work here as well, since clients should be able to understand those types once they're settled.
+```js
+{
+	// The SHA256 hash of the remote content
+	checksum: string,
+	
+	// A 32 byte hex string for decrypting the remote content payload
+	secret: string,
+	
+	// A hex string for the salt used to encrypt the remote content payload
+	salt: string,
+	
+	// A hex string for the nonce used to encrypt the remote content payload
+	nonce: string
+}
+```
+
+The content of the encoded message is a URL that points to an encrypted `EncodedContent` object. The `EncodedContent`'s content type MUST not be another `RemoteAttachment`.
+
+By using `EncodedMessage`, we can make it easier for clients to support any message content already used on the network (with the exception of `RemoteAttachment` as mentioned above).
+
+The reference implementation uses the `Attachment` type from XIP 15, but if we introduce richer types for things like images or video, those would work here as well, since clients should be able to understand those types once they're settled.
 
 ## Backward compatibility
 
@@ -40,10 +60,12 @@ Clients encountering messages of this type must already be able to deal with mes
 
 ## Reference implementation
 
-- [Implementation reference](https://github.com/xmtp/xmtp-ios/pull/68)
-- [Client Usage reference](https://github.com/xmtp-labs/xmtp-inbox-ios/pull/83)
+- [Implementation reference](https://github.com/xmtp/xmtp-ios/pull/68) (needs update)
+- [Client Usage reference](https://github.com/xmtp-labs/xmtp-inbox-ios/pull/83) (needs update)
 
 ## Security considerations
+
+Making requests to servers outside the network could reveal information similar to tracking pixels. This could be somewhat mitigated by not loading this content by default, or at least providing users with a setting.
 
 Having arbitrary data anywhere can be risky, but this is already the case for our messages, since there's no server side validation of message contents (besides size). The same protections we have now would be in place while the same pitfalls we have would still be there as well.
 
--- a/XIPs/xip-17-remote-attachment-content-type-proposal.md
+++ b/XIPs/xip-17-remote-attachment-content-type-proposal.md
@@ -24,7 +24,7 @@ Proposed content type:
 ```js
 {
   authorityId: "xmtp.org"
-  typeId: "remoteAttachment"
+  typeId: "remoteStaticContent"
   versionMajor: 0
   versionMinor: 1
 }
@@ -35,7 +35,7 @@ The encoded content MUST have the following parameters:
 ```js
 {
 	// The SHA256 hash of the remote content
-	checksum: string,
+	contentDigest: string,
 	
 	// A 32 byte hex string for decrypting the remote content payload
 	secret: string,
--- a/XIPs/xip-17-remote-attachment-content-type-proposal.md
+++ b/XIPs/xip-17-remote-attachment-content-type-proposal.md
@@ -67,7 +67,7 @@ Clients encountering messages of this type must already be able to deal with mes
 
 Making requests to servers outside the network could reveal information similar to tracking pixels. This could be somewhat mitigated by not loading this content by default, or at least providing users with a setting.
 
-Having arbitrary data anywhere can be risky, but this is already the case for our messages, since there's no server side validation of message contents (besides size). The same protections we have now would be in place while the same pitfalls we have would still be there as well.
+Having arbitrary data anywhere can be risky, but this is already the case for our messages, since there's no server side validation of message contents (besides size).
 
 ## Copyright
 
--- a/XIPs/xip-17-remote-attachment-content-type-proposal.md
+++ b/XIPs/xip-17-remote-attachment-content-type-proposal.md
@@ -1,0 +1,52 @@
+---
+xip: 17
+title: Remote Attachment Content Type
+description: Content type for remote attachments
+author: Pat Nakajima
+status: Review
+type: Standards Track
+category: XRC
+created: 2023-02-15
+---
+
+## Abstract
+
+This XRC proposes a new content type for messages that do not consist of text and/or are too large to be sent on the XMTP network. It builds on [XIP 15](https://github.com/xmtp/XIPs/pull/15)'s basic attachment type, but instead of messages being sent on the network, the message contains a URL. The contents of the URL is an `Envelope` that can be decoded by clients.
+
+## Motivation
+
+The current 1MB limit for messages on the XMTP network is too small for many types of content. To enable larger files, external storage is going to be necessary.
+
+## Specification
+
+Proposed content type:
+
+```js
+{
+  authorityId: "xmtp.org"
+  typeId: "remoteAttachment"
+  versionMajor: 0
+  versionMinor: 1
+}
+```
+
+The content of the encoded message is a URL that points to encoded Envelope data that would not be limited by the 1MB limit, since it's not being sent on the network. By using the existing `Envelope` format, we can still ensure E2E encryption for this data.
+
+The envelope itself can contain any type of message already allowed on the network. The reference implementation uses the `Attachment` type from XIP 15, but if we introduce richer types for things like images or video, those would work here as well, since clients should be able to understand those types once they're settled.
+
+## Backward compatibility
+
+Clients encountering messages of this type must already be able to deal with messages of an unknown content type, so whatever considerations they're making there should work here too.
+
+## Reference implementation
+
+- [Implementation reference](https://github.com/xmtp/xmtp-ios/pull/68)
+- Client Usage reference (TODO if this is a direction we're interested in pursuing)
+
+## Security considerations
+
+Having arbitrary data anywhere can be risky, but this is already the case for our messages, since there's no server side validation of message contents (besides size). The same protections we have now would be in place while the same pitfalls we have would still be there as well.
+
+## Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
--- a/XIPs/xip-17-remote-attachment-content-type-proposal.md
+++ b/XIPs/xip-17-remote-attachment-content-type-proposal.md
@@ -34,24 +34,24 @@ The encoded content MUST have the following parameters:
 
 ```js
 {
-	// The SHA256 hash of the remote content
-	contentDigest: string,
-	
-	// A 32 byte hex string for decrypting the remote content payload
-	secret: string,
-	
-	// A hex string for the salt used to encrypt the remote content payload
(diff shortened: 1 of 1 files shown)


XIPs/xip-17-remote-attachment-content-type-proposal.md
```diff
@@ -0,0 +1,52 @@
...
+```js
+{
+  authorityId: "xmtp.org"
+  typeId: "remoteAttachment"
+  versionMajor: 0
+  versionMinor: 1
+}
+```
+
+The content of the encoded message is a URL that points to encoded Envelope data that would not be limited by the 1MB limit, since it's not being sent on the network. By using the existing `Envelope` format, we can still ensure E2E encryption for this data.
```
neekolas, line 33: Would be helpful to specify the encryption scheme. How do we pass keys to the reader? What types of keys are allowed?

XIPs/xip-17-remote-attachment-content-type-proposal.md
```diff
@@ -0,0 +1,52 @@
...
+
+The envelope itself can contain any type of message already allowed on the network. The reference implementation uses the `Attachment` type from XIP 15, but if we introduce richer types for things like images or video, those would work here as well, since clients should be able to understand those types once they're settled.
```
neekolas, line 35: How is the envelope serialized? Protobuf?

XIPs/xip-17-remote-attachment-content-type-proposal.md
nakajima: It would use the same encryption we use for publishing envelopes to the network, but instead of publishing to the network it would be to some other host.

XIPs/xip-17-remote-attachment-content-type-proposal.md
nakajima: I think protobuf works since clients already need to speak it.

XIPs/xip-17-remote-attachment-content-type-proposal.md
neekolas: Which type, though. For V1 messages we use X3DH and for V2 messages we use random `TopicKey`s. In either case, we would need to figure out where to store the salt and the nonce. Those values are unique and random, and we wouldn't want to re-use the same values that went into the message since nonce re-use is an attack vector that could lead to messages being compromised.

So, we have a few options:
1. In the message payload, include the `nonce` and `salt` and use whatever secret was used to encrypt the message. The remote content would just be the encrypted bytes of the message
2. Keep the message payload light, and just include a link to the content. The remote content would need to be formatted as `xmtp.Ciphertext` proto message. Re-use the shared secret for the message (either V1 X3DH or V2 TopicKey)
3. Use an ephemeral key. The message payload would include the key (maybe 32 bytes of randomness), nonce, and salt. The remote message would be the encrypted payload only.

Of those options (those are just the three off the top of my head), I'd lean towards 3. Gives the most security, since each payload is encrypted with a different key, and means that we don't have to decode another protobuf to read the message.

neekolas:
One thing I'm wondering is whether we actually want the remote payload to be able to be stored as a full `EncodedContent` with XIP-5 compatible content-types. That allows you to store any message compatible with XMTP remotely and clients already have a flexible set of decoding options. Of course, one of those content types could be the proposed attachment content type that relies on MIME, but you could also use other XIP-5 content types that have built in decoders.

Not sure I've fully thought through the implications, but it seems like it would let us use all the content decoding tools we already have. Just with much larger payloads.

XIPs/xip-17-remote-attachment-content-type-proposal.md
neekolas: @jazzz @richardhuaaa @michaelx11 curious what the ProtoPod thinks

nakajima: I think maybe I'm missing something with how our encryption works. My thought for the remote payload is that it's an `xmtp.Envelope` with a topic, timestamp and message. In the same way we can send those as push notifications, I was thinking we could also make them the remote payload. It was my understanding that anyone can query for those but the only way to decrypt it is if you have the conversation key material as well (my thinking was that we'd be using the V2 message format).

nakajima:
> One thing I'm wondering is whether we actually want the remote payload to be able to be stored as a full EncodedContent with XIP-5 compatible content-types.

Ya, this was my thinking.

nakajima:
Added a reference implementation example PR in the Inbox iOS app to try to show what I've got in mind here: https://github.com/xmtp-labs/xmtp-inbox-ios/pull/83

XIPs/xip-17-remote-attachment-content-type-proposal.md
neekolas: So, I guess we could do the full envelope. That's basically option 2 above. The topic and timestamp feel a little superfluous (won't they always be the same as the on-network message?), but are harmless.

nakajima: I like the full envelope because clients already need to be able to handle decoding those. I agree about the topic/timestamp being superfluous but I think it's worth it for the simplicity.

jazzz: I'd lean towards option 3. where:
- PointerMessage contains a symmetricKey and all the parameters to decrypt it.
- Server contains the content payload only.

I am weary of xmtp.Envelope as the abstraction interface. 
- All possible Envelopes stored remotely requires exponentially more testing. 
   - "What happens if a Key Revocation message is stored off network?",  
   - "what happens if an Invite is no longer accessible?"
   - "What if a remoteContent points to another removeContent?"
- Coupling these messages to a specific XMTP Message version would make future migrations more complicated.

XIPs/xip-17-remote-attachment-content-type-proposal.md
```diff
@@ -0,0 +1,52 @@
...
+Clients encountering messages of this type must already be able to deal with messages of an unknown content type, so whatever considerations they're making there should work here too.
+
+## Reference implementation
+
+- [Implementation reference](https://github.com/xmtp/xmtp-ios/pull/68)
+- [Client Usage reference](https://github.com/xmtp-labs/xmtp-inbox-ios/pull/83)
+
+## Security considerations
+
+Having arbitrary data anywhere can be risky, but this is already the case for our messages, since there's no server side validation of message contents (besides size). The same protections we have now would be in place while the same pitfalls we have would still be there as well.
```
jazzz, line 48: Is there a strategy for avoiding circular loops? If the RemoteContent contains an envelope, can it also contain a another RemoteContent object?

jazzz (commented):
As currently written this seems problematic. 

If this path is desired I would suggest
- The remote content type should point to contentbytes only, not another message.
- Server stored content should be completely independent from XMTP -- e.g. Raw Image bytes. Message Versions and content versions should be able to increment independently without consideration.
- Server stored content should be encrypted using a onetime symmetric key.
- RemoteContent type should contain a digest of the expected stored content.
- RemoteContent should require a user action prior to fetching from server

XIPs/xip-17-remote-attachment-content-type-proposal.md
neekolas: Ooh, that's a very good point. Would be an easy DOS vector against clients if protections aren't in place.

nakajima:
@jazzz Thanks for the feedback! I'll take another pass at this and hopefully address all your concerns.

XIPs/xip-17-remote-attachment-content-type-proposal.md
neekolas: I hear you @jazzz on the potential complexity of having the remote payloads be some sort of Content Type. I'm unsold as `xmtp.Envelope` being the right wrapper. But an encrypted `EncodedContent` payload will make things a lot more usable for developers than a pile of unstructured bytes. Especially if these are messages that are meant to be displayed inline, where we would want to have the regular set of tools for decoding available that we offer for regular messages.

Otherwise you are going to have one flow to handle remote images, and a totally different decoding flow for on-network images. That seems less than ideal.

> "What happens if a Key Revocation message is stored off network?"

Since we don't have Key Revocation messages yet, hard for me to answer conclusively. Does seem like we may want certain meta-messages to live outside the normal message encoding flow and be typed based on contentTopic. We do this with `Invites` currently, which are not a `ContentType` at all.

> "what happens if an Invite is no longer accessible?"

Are you asking what happens if someone sends a conversation invite as a remote message? That should not be possible, as invites are separate from messages and are not `EncodedContent` or sent in message channels.

> "What if a remoteContent points to another removeContent?"

Both the spec and the example implementation need a mitigation for this kind of attack. Seems like we can have some sort of recursion limit in the SDK and throw an error after N loops.

nakajima:
Updated the proposal to go with [option 3](https://github.com/xmtp/XIPs/pull/18#discussion_r1112333490), adding a required `checksum` field as well. I also added a note about potential pixel tracker concerns.

XIPs/xip-17-remote-attachment-content-type-proposal.md
jazzz: > But an encrypted EncodedContent payload will make things a lot more usable for developers than a pile of unstructured bytes.

I'll start by saying I can get behind this. EncodedContent payload seems like the an acceptable place an alevates many of my concerns.

-------------

To round out my thoughts: If the xmtp.Envelope wrapper is excluded then I see two options:

remoteContent

*OptionA* -- Making remoteContent a composable type Leads to maximum utility. Any type can now be stored remotely with no changes required.

*OptionB* -- Making RemoteContent an implementation pattern for other type means it could be handled on a type by type basis.  Larger types (Images, Video, File) could support remoteContent but this is not required of all types.

My argument for B is not particularly strong: There are only few types that could really benefit from OptionA, yet potential remote storage may become a burden for future ContentType Developers . Any future ContentTypes will now need to be evaluated against a wider threat model, as independent ContentServers are outside of the protection of the Network consensus protocol. I don't have any hard examples as there are few formal XIP ContentTypes defined. Given this Option A is compelling.

jazzz: >We do this with Invites currently, which are not a ContentType at all.

Great point.

> That should not be possible

Great point. Despite the payload being able to hold an invite, it would not be possible to get an EncodedContent payload into the invite processing flow.

XIPs/xip-17-remote-attachment-content-type-proposal.md
```diff
@@ -0,0 +1,74 @@
...
+  versionMinor: 1
+}
+```
+
+The encoded content MUST have the following parameters:
+
+```js
+{
+	// The SHA256 hash of the remote content
+	checksum: string,
```
jazzz, line 38: checkSums are usually used in the context of data corruption. In this context the goal is message Integrity: Can we use "contentDigest" (or something similar) to reflect that?

nakajima: Yep!

nakajima: Is SHA256 an ok hash to use?

XIPs/xip-17-remote-attachment-content-type-proposal.md
```diff
@@ -0,0 +1,74 @@
...
+The current 1MB limit for messages on the XMTP network is too small for many types of content. To enable larger files, external storage is going to be necessary.
+
+## Specification
+
+Proposed content type:
+
+```js
+{
+  authorityId: "xmtp.org"
+  typeId: "remoteAttachment"
```
jazzz, line 27: Thoughts on limiting this specification to the scope of StaticContent only?

Currently it is ambiguous whether the ContentServers can return Dynamic content. By explicitly only supporting static content then caching and Post-Delivery-storage approaches can avoid complications from cache Invalidation etc. 

Adding of the `contentDigest` mostly achieves this anyways, being explicit would make things cleaner

XIPs/xip-17-remote-attachment-content-type-proposal.md
snormore: Could be a good use-case for a CID from https://github.com/multiformats/js-multiformats (similar to what IPFS is using)

XIPs/xip-17-remote-attachment-content-type-proposal.md
nakajima: What do you think of maybe `remoteStaticContent`?

XIPs/xip-17-remote-attachment-content-type-proposal.md
nakajima: Curious since I'm not super familiar with IPFS, what would that CIDs get us that SHA256 doesn't?

snormore: Multihash/CID is kind of just an abstraction around different hashing methods for content. This is how IPFS docs describe it: https://docs.ipfs.tech/concepts/content-addressing/

Screenshot 2023-02-21 at 2 23 42 PM

michaelx11:
Jumping in on this a bit late with some dumb questions. My personal focus when evaluating changes like these is always on 1) clarity and 2) easily analyzable security properties.

In that sense, two questions:
1) If we just bump our message limit up to 10MB, do we need a remote content type at all for our current needs i.e. small videos, photos, etc?
2) If we separate the RemoteContent concept into its own thing, I think we run the risk of ambiguity about how to fetch things. In my mind, the most unambiguous way to handle this is to have a new Content-Type for each remote use case.

Example:
- For webm videos meant to be fetched over HTTPS, we can have an `xmtp.org/https_webm:1.0`
- For webm videos meant to be fetched over IPFS: `xmtp.org/ipfs_webm:1.0`

The idea here is that given just a content-type, a developer knows exactly how to fetch and interpret the content. Additionally, organizations can allowlist content-types if there are concerns around remote content (e.g. IP leakage, read receipt leakage, malicious payload vectors). Unifying at the content-type level will avoid folks having to do logic like "allow remote content, but only for these content-types, and only fetched in these ways". The logic would instead be "if content type not in this allowlist, do not process"

nakajima:
> If we just bump our message limit up to 10MB, do we need a remote content type at all for our current needs i.e. small videos, photos, etc?

I think partially it does, but then clients need to contend with huge messages being fetched and decoded unless they use a separate topic for these types of things.

> If we separate the RemoteContent concept into its own thing, I think we run the risk of ambiguity about how to fetch things. In my mind, the most unambiguous way to handle this is to have a new Content-Type for each remote use case.

I think maybe the case could be made for separate remote content types based off how they're fetched, like one for HTTP vs one for IPFS. I still think the remote payloads being `EncodedContent` keeps things simplest for clients though. We could also add a content type field to the metadata that gets sent and ensure it matches that of the remote `EncodedContent`. Clients could use this to ignore remote messages that they aren't cool with.

nakajima:
Thinking about it some more, if an organization ever did want to block certain types of content, they could just block all `remoteStaticContent` messages and if they still had a need for remote content, they could just make their own content type.

nakajima:
SDK implementation updated: https://github.com/xmtp/xmtp-ios/pull/68
Client usage updated: https://github.com/xmtp-labs/xmtp-inbox-ios/pull/83

Note that neither are particularly production ready yet (need to handle a bunch more cases) but I was able to get messages sending with encrypted remote payloads and receive them in the app:

Screenshot of xmtp inbox app with a remote image attached

nakajima:
Are there any more pressing security concerns with this approach or would it be alright to ship and learn? Just trying to avoid making perfect the enemy of the good while still keeping things safe for folks.

XIPs/xip-17-remote-attachment-content-type-proposal.md
michaelx11: I'm with @jazzz in that I'm hesitant to define a `RemoteContent` type that can wrap any normal `XMTP` content type. I worry about the possible attack surface caused by remote fetches such as SSRF (e.g. attacker puts 127.0.0.1:[some important port]/[some sensitive path]). We'd have to consider the cross product of `All RemoteContent actions X All ContentTypes`. I'd be more comfortable if we keep `RemoteContent`'s capabilities very well described and limited. For example: must be a GET over HTTPS.

michaelx11:
I'm okay with shipping the proposal but with a few fixed URL schemes to start, such as `https://`. We should explicitly state this in the proto/spec if possible, since schemes like `file://` should be absolutely not allowed. Any time code is processing an untrusted URL there is a rich world of possible attacks. We can do our developers a favor by making these things (like expected scheme) an explicit part of the protocol.

nakajima:
@michaelx11 What do you think of making the scheme part of the parameters? To start the only valid value would be `https` and SDKs could ensure that URLs only start with that. Later if we wanted to add IPFS in a newer version of this content type, we could put it there?

michaelx11:
Sounds good to me! We can make it an enum and add more supported schemes as we go

nakajima:
Rad, I updated the proposal to include the scheme parameter as well as a note that the content must be acccessed via GET request.

nakajima:
The reference PRs have been updated to include the mandatory scheme parameter, enforcing `https://` and `GET` requests only.

nakajima:
>> If we just bump our message limit up to 10MB, do we need a remote content type at all for our current needs i.e. small videos, photos, etc?

> I think partially it does, but then clients need to contend with huge messages being fetched and decoded unless they use a separate topic for these types of things.

Given @nmalzieu's comment [here](https://github.com/xmtp/XIPs/pull/15#discussion_r1114178027), I actually don't think bumping the message limit to 10 mb is the best course of action anymore.

neekolas:
I still don’t love the fact that we have a bunch of developers wanting to build custom content types, and our answer is to circumvent the content type system altogether. It feels like it shifts a lot of work onto the developer to figure out which MIME types they support, and how to handle the ones they don't. 

If we had specific content types for each type of attachment as @michaelx11 suggested, we can take a lot of that burden off the developers plate. Developers are either committing to support a given type or not.

neekolas:
I like where this landed

neekolas (approved):

XIPs/xip-17-remote-attachment-content-type-proposal.md
```diff
@@ -0,0 +1,52 @@
...
+## Reference implementation
+
+- [Implementation reference](https://github.com/xmtp/xmtp-ios/pull/68)
+- [Client Usage reference](https://github.com/xmtp-labs/xmtp-inbox-ios/pull/83)
+
+## Security considerations
+
+Having arbitrary data anywhere can be risky, but this is already the case for our messages, since there's no server side validation of message contents (besides size). The same protections we have now would be in place while the same pitfalls we have would still be there as well.
+
+## Copyright
```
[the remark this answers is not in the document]
neekolas: I think we should strongly encourage developers using this content type to implement click to load. 

Easy to imagine someone sending 100 messages with 1gb remote attachments as an attack.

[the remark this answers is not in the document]
neekolas: At the app level you could then store an "always download attachments from this sender" preference

XIPs/xip-17-remote-attachment-content-type-proposal.md
```diff
@@ -0,0 +1,77 @@
...
+  // A hex string for the nonce used to encrypt the remote content payload
+  nonce: string,
+  
+  // The scheme of the URL. Must be "https://"
+  scheme: "https://"
+}
+```
+
+The content of the encoded message is a URL that points to an encrypted `EncodedContent` object. The content must be accessible by an HTTP `GET` request to the URL. The `EncodedContent`'s content type MUST not be another `RemoteAttachment`.
+
```
jazzz, line 55: > content type MUST not be another `RemoteAttachment`.

I am glad to see this here.

jazzz (approved):
@nakajima great job incorporating all the feedback on this. I think this has ended up in a good place. 

Excited to see what developers do with this new power

🚢

nakajima:
Thanks everyone for your feedback!

PR #23, opened 2023-03-24
Reaction content type proposal

Im-Kunal-13:
* Create xip-reaction-content-type.md

### DIFF
--- a/XIPs/xip-reaction-content-type.md
+++ b/XIPs/xip-reaction-content-type.md
@@ -32,24 +32,22 @@ Proposed content type:
 
 The reaction MUST include the following parameters:
 
-Adding info about the message we’re reacting to (it’s message id, sender address, message content, content type id and the reaction). 
-
 ```ts
 {
-    messageId: string
-    senderAddress: string
-    messageContent: string
-    contentTypeId: string
-    reaction: string
+    // The message ID that this reaction is being added to
+    reactingToID: string
 }
 ```
 
-The content of the encoded message is arbitrary data. It's up to clients to use the messageId , senderAddress , messageContent, contentTypeId and reaction to determine how to render the message.
+The content of the encoded message must be a string consisting of a single emoji character.
 
 ## Backward compatibility
 
 Clients encountering messages of this type must already be able to deal with messages of an unknown content type, so whatever considerations they're making there should work here too.
 
+## Reference implementation
+
+- [PR implementation for the JS SDK](https://github.com/xmtp/xmtp-js-content-types/blob/d859187c00d216069cffbb2f6a847cd276e5225a/reaction/src/Reaction.ts)
 
 ## Security considerations
 
--- a/XIPs/xip-reaction-content-type.md
+++ b/XIPs/xip-reaction-content-type.md
@@ -1,0 +1,60 @@
+---
+xip: 
+title: Reaction Content Type
+description: Content type for reactions
+author: Kunal Mondal
+status: Draft 
+type: Standards Track
+category: XRC
+created: 2023-03-24
+---
+
+## Abstract
+
+This XRC proposes a new content type for messages that enriches the messaging experience and let's users express themselves using emojis. It attempts to include only the bare minimum for clients to determine how to display such messages in order to provide flexibility for more rich types in the future. 
+
+## Motivation
+
+When it comes to having a comfortable conversation, reacting to messages with emojis is necessary as it gives users the freedom to express their messages in form of emoji. In order to provide users a real messaging platform, XMTP needs to provide this capability.
+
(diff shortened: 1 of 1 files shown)


XIPs/xip-reaction-content-type.md
```diff
@@ -0,0 +1,60 @@
...
+
+The reaction MUST include the following parameters:
+
+Adding info about the message we’re reacting to (it’s message id, sender address, message content, content type id and the reaction). 
+
+```ts
+{
+    messageId: string
+    senderAddress: string
+    messageContent: string
```
neekolas, line 41: Would this be the `message.content` from the message you are reacting to? Why do we need to duplicate that in the Reaction? Couldn't clients just read the original message?

XIPs/xip-reaction-content-type.md
```diff
@@ -0,0 +1,60 @@
...
+    contentTypeId: string
+    reaction: string
```
neekolas, line 43: If the reaction field is only ever expected to be a single UTF-8 emoji character, we can indicate and enforce that the maximum allowed length for this field is 4 bytes. Otherwise someone can dump a short novel into the field and it would be up to the receiving client to figure out what to do with that.

XIPs/xip-reaction-content-type.md
Im-Kunal-13: yes `messageContent` is the `message.content` from the message we're replying to. 

It's necessary to duplicate it for the following use cases : 

1. We're loading the last messages of all the conversations of a user to show them in the chat previews. Whenever the last message is a Reaction, we need the `messageContent` and `contentTypeId` to display which message it has been reacted to. ( See the image)

image

2. There are cases where the message we reacted to is not loaded yet (due to the pagination) and so we won't have `messageContent` and `contentTypeId` info if we try to filter and find it from the message's state in the application through the Reaction's messageId.

XIPs/xip-reaction-content-type.md
Im-Kunal-13: Yes exactly that's better. At what step do you recommend the validation should be done ?

XIPs/xip-reaction-content-type.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+The reaction MUST include the following parameters:
+
+Adding info about the message we’re reacting to (it’s message id, sender address, message content, content type id and the reaction). 
+
+```ts
+{
+    messageId: string
+    senderAddress: string
```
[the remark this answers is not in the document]
nakajima: > Yes right. We can use the senderAddress from the XMTP message itself instead of relying on this. Should I remove it and commit ?

Ya you can remove this and commit.

XIPs/xip-reaction-content-type.md
nakajima: I think we should just pass IDs here, since duplication could lead to different content being passed in different places which would lead to inconsistency. It also doesn't account for non-text messages.

Im-Kunal-13: Yes we can remove the `messageContent `and `contentTypeId` and only keep the `messageId` and `reaction` for this to work.

Although as we won't have the the content of the message we reacted to, we can't show them in the preview. It will be just the reaction.

image

XIPs/xip-reaction-content-type.md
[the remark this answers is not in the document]
Im-Kunal-13: > > Yes right. We can use the senderAddress from the XMTP message itself instead of relying on this. Should I remove it and commit ?
> 
> Ya you can remove this and commit.

Okay sure.

nakajima:
Just pushed some updates here to remove duplicate fields from the proposal as well as tweak where the reaction actually gets stored (the content rather than as a parameter).

I've also added a link to a reference implementation here.

nplasterer:
I just wanted to note that we added reaction content types into the JS SDK here: https://github.com/xmtp/xmtp-js-content-types/pull/13 & https://github.com/xmtp/xmtp-js-content-types/pull/9 and we gave a reference implementation in our React playground here: https://github.com/xmtp/xmtp-react-playground/pull/1
This should unblock any jsSDK or reactnativeSDK integrators.
We have tasks to add support to kotlin, swift, and flutter soon.

jhaaaa:
After more than six months of inactivity on this XIP, we have set its status to Stagnant. We have [merged the XIP to the repo](https://github.com/xmtp/XIPs/blob/main/XIPs/z-stagnant_xip-23-reaction-content-type.md) for historical purposes and are closing this PR. Thank you very much for your contribution, @Im-Kunal-13!

PR #25, opened 2023-04-05
Payment content type XIP

aeyshubh:
The Xip can be used for sending payments through chats.


mohitthakkar30 (commented):
Cool

AmoghVaishnav:
Looks like a really cool xip ,easy to understand and excited to use it in my app.

PR #28, opened 2023-04-13
Universal Consent Layer

polmaire:
Hello everyone, please find attached a proposition to add a layer of consent to the XMTP protocol. The goal is this document is to propose a way to make user consent interoperable across XMTP clients (for example, if I block someone on a client, this user is also blocked on other clients). Thank you in advance for reviewing this proposition!

### DIFF
--- a/XIPs/xip-universal-consent-layer.md
+++ b/XIPs/xip-universal-consent-layer.md
@@ -1,5 +1,4 @@
 # XIP-XX: Universal Consent Layer for XMTP
-
 | XIP                |                                            |
 |--------------------|--------------------------------------------|
 | Title              | Universal Consent Layer for XMTP           |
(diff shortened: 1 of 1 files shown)


XIPs/xip-universal-consent-layer.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Motivation
+
+Currently, the XMTP network allows anyone to write to anyone without the option to opt-in or opt-out of conversations with other users. Implementing a universal consent layer will provide users with more control over their communication preferences, reduce spam, and enhance privacy.
+
+## Specification
+
+### Consent layers
+
+1. **Opted-in**: The user has either sent at least one message in a conversation with the other user or explicitly accepted talking with the other user.
```
neekolas, line 23: One concern here is the number of messages needed to determine whether the user has sent a message in the convo. If clients are expected to read through the message history of every conversation, that is a lot of overhead (particularly for mobile apps).

If the most recent messages in the convo are all from the other party, the client would have to read through all of them until they found one from the user.

XIPs/xip-universal-consent-layer.md
```diff
@@ -0,0 +1,55 @@
...
+3. **Neither opted-in nor opted-out**: The user received a message from another user but did not send anything nor clicked on any consent button.
+
+### User Interface
+
+- Clients should create a "requests" section in their user interface containing all conversations with users in the "Neither opted-in or opted-out" category.
+- Clients can add an additional layer of consent, "Implicitly opted-in", at the client level to automatically accept users if there is a good chance of a positive interaction (e.g., having an on-chain transaction with the user or following the user on Lens or Farcaster). This additional layer is not at the protocol level for now.
+
+### User Settings
+
+- User settings do not currently exist at the protocol level. There should be another XIP to address this specific topic.
```
neekolas, line 34: One concern we are working through with protocol level user settings is trust. While the settings can be encrypted/signed with the user's keys, if we use the XMTP keys to authenticate the settings it means that a malicious (or growth-minded) client application could modify those settings without the user's knowledge.

For example, apps might "follow" their own support channels in the background, or even sell opt-ins to marketers.

Other, non-malicious, apps would then see these settings as authentic.

The only way around this problem I can see is to require these kinds of setting changes to be signed by the wallet, which does require explicit user consent...but also adds significant friction.

XIPs/xip-universal-consent-layer.md
nmalzieu: @neekolas 100% agree!
We propose that this first version of the XIP only deals with explicit opt-in and opt-out.
Let's remove implicit opt-in (which is hard for the performance reason you just mentioned) 
We'll also add a section describing how an implicit opt-in should look like in clients that want to implement it (i.e. clients that sync all messages)

jhaaaa:
After more than six months of inactivity on this XIP, we have set its status to Stagnant. We have [merged the XIP to the repo](https://github.com/xmtp/XIPs/blob/main/XIPs/z-stagnant_xip-28-universal-consent-layer.md) for historical purposes and are closing this PR.

Thank you very much for your meaningful contribution which inspired [XIP-42: Universal allow and block preferences](https://github.com/xmtp/XIPs/blob/main/XIPs/xip-42-universal-allow-block-preferences.md), @polmaire!

PR #31, opened 2023-09-11
Create Message Backup Requests XIP

neekolas:
An opinionated proposal for Portable Inbox on top of XMTP V3.

### DIFF
--- a/XIPs/xip-31-message-backup-request-system.md
+++ b/XIPs/xip-31-message-backup-request-system.md
@@ -15,7 +15,7 @@ A proposed system for transferring backups of message history between installati
 
 ## Motivation
 
-The current [XMTP V3 Protocol Specification](https://github.com/xmtp/libxmtp) creates a new XMTP identity for each installation associated with a blockchain account, instead of sharing a single set of keys across all installations like V1/V2. New installations will receive all messages sent to a blockchain account after they are created, but V3 does _not_ specify how to receive messages sent before installation creation. Having conversation history sync between devices is a popular feature of V1 and V2 of the XMTP protocol and there is clear demand from developers for this to be possible as part of V3.
+The [XMTP V3 Protocol Specification](https://github.com/xmtp/libxmtp) creates a new XMTP identity for each installation associated with a blockchain account, instead of sharing a single set of keys across all installations like V1/V2. New installations will receive all messages sent to a blockchain account after they are created, but V3 does _not_ specify how to receive messages sent before installation creation. Having conversation history sync between devices is a popular feature of V1 and V2 of the XMTP protocol and there is clear demand from developers for this to be possible as part of V3.
 
 There are a number of cases where message history synchronization is desirable:
 
@@ -37,7 +37,7 @@ It's also worth noting that there are many use-cases for XMTP where message hist
 1. Message history synchronization should be opt-in for developers.
 1. It should be simple for developers to implement message history synchronization correctly. Implementation work should be focused on application-specific UX choices, with the SDK responsible for core transport and security decisions.
 1. Compatibility between applications is expected. Applications should be able to share history irrespective of operating system, SDK version, or device type (web vs mobile vs server).
-1. Even if developers of some popular applications choose to not support message history synchronization, it should be simple for users to opt themselves in.
+1. Even if developers of some popular applications choose to not support message history synchronization, it should be simple for users to opt enable themselves.
 
 ## Specification
 
--- a/XIPs/xip-31-message-backup-request-system.md
+++ b/XIPs/xip-31-message-backup-request-system.md
@@ -199,7 +199,7 @@ I am also proposing that XMTP Labs runs a Backup Storage Provider as a free publ
 
 #### [Mesage Protos](https://github.com/xmtp/proto/blob/xmtpv3/proto/v3/message_contents/message.proto)
 
-```proto3
+```proto
 // The decrypted message contents of any message on the installation's messaging topic
 message PadlockMessagePayload {
     EdDsaSignature header_signature = 1;
@@ -234,7 +234,7 @@ message DirectMessage {
 
 #### [Contact protos](https://github.com/xmtp/proto/blob/xmtpv3/proto/v3/message_contents/public_key.proto)
 
-```proto3
+```proto
 enum ContactRole {
     UNKNOWN_CONTACT_ROLE = 0;
     FULL = 1;
--- a/XIPs/xip-31-message-backup-request-system.md
+++ b/XIPs/xip-31-message-backup-request-system.md
@@ -73,7 +73,9 @@ class S["Backup Storage Provider"]{
 
 ### Remote Message Backups
 
-Remote message backups work by Backup Requesters sending a special message type (`MessageHistoryBackupRequest`) across the XMTP network to all other installations signed by the same blockchain account as the sender (Message Backup Providers). Upon receipt of these messages, Message Backup Providers should display a prompt to the user asking whether they consent to share their message history with the requesting installation. Upon approval, the application will convert their local database into a standard Message History Backup File and upload it to the Backup Storage Provider specified in the `MessageHistoryBackupRequest`.
+Remote message backups work by Backup Requesters sending a special message type (`MessageHistoryBackupRequest`) across the XMTP network to all other installations signed by the same blockchain account as the sender (Message Backup Providers). Upon receipt of these messages, Message Backup Providers should display a prompt to the user asking whether they consent to share their message history with the requesting installation. The prompt should display the Verification PIN from the request - which will also be displayed in the Backup Requester's app - to ensure that the user is approving the correct request.
+
+Upon approval, the application will convert their local database into a standard Message History Backup File, encrypt it, and upload it to the Backup Storage Provider specified in the `MessageHistoryBackupRequest`.
 
 For mobile applications already handling push notifications, `MessageHistoryBackupRequest`s would become a special case of push notification handling.
 
--- a/XIPs/xip-31-message-backup-request-system.md
+++ b/XIPs/xip-31-message-backup-request-system.md
@@ -443,7 +443,7 @@ TODO
 
 ### Risks and drawbacks to Remote Message Backups
 
-- Developers of the Message Backup Provider could implement the consent flow in a way that confuses users into accepting backup requests from malicious applications. PIN verification may not be implemnted across all providers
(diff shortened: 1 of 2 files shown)


XIPs/xip-31-message-backup-request-system.md
```diff
@@ -0,0 +1,466 @@
...
+1. Message history synchronization should be opt-in for developers.
+1. It should be simple for developers to implement message history synchronization correctly. Implementation work should be focused on application-specific UX choices, with the SDK responsible for core transport and security decisions.
+1. Compatibility between applications is expected. Applications should be able to share history irrespective of operating system, SDK version, or device type (web vs mobile vs server).
+1. Even if developers of some popular applications choose to not support message history synchronization, it should be simple for users to opt enable themselves.
+
+## Specification
+
+This specification offers the user two modes of message backup, which accomodate different use-cases: Remote Message Backups and Backup Account Files. Remote Message Backups are the lowest friction solution, provided the blockchain account has access to another existing installation. Backup Account Files are an emergency solution for cases where the user has lost access to all of their XMTP applications (for example, if they lost the only device they have used to connect to XMTP).
```
bwcDvorak, line 44: I may be splitting hairs, but some social media users have preconceptions about the term "backup account". Suggest we consider naming our 2nd use case "Account Backup File" to avoid concept collision with the above.

neekolas: I worry "Account Backup File" might lead developers astray. This process is creating a secondary XMTP account just for backups, and then saving it to a file. When I hear "Account Backup File" it's easy to assume that it is a backup of your regular account.

We can workshop the name a bit. I'm not married to "Backup Account File".

XIPs/xip-31-message-backup-request-system.md
```diff
@@ -0,0 +1,466 @@
...
+    B->>S: Request file from URL in response
+    S-->>B: Receive file
+    B-->>B: Decrypt and load file into database
+```
+
+### Backup Account Files
+
+Backup account files are designed as a solution for users who have lost access to all of their XMTP applications. To create a Backup Account File, an application will generate a new set of XMTP keys and register the installation on the XMTP network. The installation contact published to the network will be set with a type of `backup`, which instructs all well-behaved clients to disregard any messages from this installation other than `MessageHistoryBackupResponse`s. This means that a compromised backup installation cannot be used to send messages on the user's behalf, which makes these files a less valuable target for attack.
+
+Once a backup account's contact has been published to the network, their installation messaging topic will receive a copy of all new messages sent to the associated blockchain account.
```
richardhuaaa, line 140: One thing to mention is that the backup client doesn't have access to historical messages. We would have to introduce new payload formats to retransmit all existing history to the backup client over the XMTP network

XIPs/xip-31-message-backup-request-system.md
```diff
@@ -0,0 +1,466 @@
...
+
+TODO
+
+## Security considerations
+
+### Risks and drawbacks to Remote Message Backups
+
+- Developers of the Message Backup Provider could implement the consent flow in a way that confuses users into accepting backup requests from malicious applications. PIN verification may not be implemented across all providers
+- A malicious Message Backup Provider could choose to censor messages in the backup, or add spoofed messages, since the messages in the backup are unauthenticated.
+- Bugs in the validation of requests in Message Backup Providers would be very very bad. This would need to be very well tested code, since a compromise here would effectively be a 0-click exploit of someone's XMTP account.
```
richardhuaaa, line 450: Given the sensitivity, it might make sense to structure this code as a separate crate/extension/package, to clearly separate it from the mainline protocol, but included as a default. This would allow people to examine it with a higher degree of scrutiny, as well as for extremely sensitive apps to strip it at compile time.

XIPs/xip-31-message-backup-request-system.md
```diff
@@ -0,0 +1,466 @@
...
+## Reference implementation
+
+TODO
+
+## Security considerations
+
+### Risks and drawbacks to Remote Message Backups
+
+- Developers of the Message Backup Provider could implement the consent flow in a way that confuses users into accepting backup requests from malicious applications. PIN verification may not be implemented across all providers
+- A malicious Message Backup Provider could choose to censor messages in the backup, or add spoofed messages, since the messages in the backup are unauthenticated.
```
richardhuaaa, line 449: If we want to add defense in depth:

1) We could consider requiring a wallet signature in the backup provider before sending the backup response. This allows the application layer to prove to the SDK layer that the correct information (including PIN) was surfaced to the user and consent was provided, minimizing implementation error, and requiring less context from app developers. In the current design, the security of remote backups is only as strong as the weakest app in the system.

2) We could consider requiring a wallet signature in the backup requester (optionally rolled into the same signature for installation key provisioning), to be validated on the provider before surfacing the consent flow. A particular area of potential weakness is the behavior with revoked installations. With the current design, any bug or momentary lapse in revocation logic (for example if a revocation payload was hidden from an installation for a brief moment in time) could allow a revoked installation to issue backup requests, or allow backup spoofing from a revoked installation. Additionally, in some cases a user may not actually want their message history on a new app, but there is nothing stopping a backup request from being issued regardless.

Arguable if these are worth the extra friction, mentioning here to provoke discussion

richardhuaaa: It would also be nice to restrict the situations in which a backup can be requested - ideally only when creating a new installation, and at no other time.

richardhuaaa (commented):
I love how comprehensive this is - left some initial comments for discussion

XIPs/xip-31-message-backup-request-system.md
neekolas: > We could consider requiring a wallet signature in the backup provider before sending the backup response. This allows the application layer to prove to the SDK layer that the correct information (including PIN) was surfaced to the user and consent was provided

While I guess it would lower the risk of developer error, the backup provider by definition already has all the messages. Why would they go through the trouble of using the backup service to exfiltrate messages without user consent when they can just use the SDK directly to get them? 

> A particular area of potential weakness is the behavior with revoked installations. With the current design, any bug or momentary lapse in revocation logic (for example if a revocation payload was hidden from an installation for a brief moment in time) could allow a revoked installation to issue backup requests, or allow backup spoofing from a revoked installation.

I'd like to go deeper on what the concerns with revocation are. It seems like we should address those head-on before we make any decisions that try to compensate for perceived weaknesses. Having access to the installation keys should be a strong assurance that the installation is controlled by the wallet owner given we are trusting this for message authentication. As part of the implementation we can (and should) force the Backup Provider to refresh the wallet address's contact topic to see if there are any recent revocation messages.

> Additionally, in some cases a user may not actually want their message history on a new app, but there is nothing stopping a backup request from being issued regardless.

I wonder if we can handle this point through strong UX guidance, rather than relying on wallet signatures. I worry that if we have too many wallet signatures as part of sign-in we will induce signature-blindness in the user, risking them clicking through signature dialogs carelessly. I do see the benefit in making backup requests an explicit user action, rather than an implicit part of sign-up. It could be annoying to receive a bunch of push notifications any time you access an XMTP app. Especially if apps don't explain what is happening to the user clearly.

> It would also be nice to restrict the situations in which a backup can be requested - ideally only when creating a new installation, and at no other time.

I like this. We should figure out how to enforce it, but it's a sensible idea.

XIPs/xip-31-message-backup-request-system.md
neekolas: That's true. It will only have a complete history of your XMTP messaging if it is created as part of your first interaction with the XMTP network. If you create it later, it will only have history starting at the time of it's creation.

XIPs/xip-31-message-backup-request-system.md
```diff
@@ -0,0 +1,466 @@
...
+
+Once a backup account's contact has been published to the network, their installation messaging topic will receive a copy of all new messages sent to the associated blockchain account.
+
+The newly generated XMTP keys are encrypted and downloaded to the user's device. They may be encrypted by either a passphrase or a wallet signature. Users or applications may decide to store the Backup Account File in a private cloud storage provider (iCloud, Google Drive, etc). Or the file could simply be stored on-device with the risk of loss if the device was inaccessible.
```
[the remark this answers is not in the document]
neekolas: > The minimal friction case is storing the backup on the same device as the app - but this would not be accessible either if the device is lost. On the flip-side, the cloud hosting case requires significant user or developer friction, as a preventive measure for a somewhat rare scenario.

In the case of mobile apps, storing the data locally and persisting to the cloud can be very similar. iOS and Android both give app developers built-in cloud-synced storage to work with. The lift might not be so bad.

> For those truly fearful of this scenario and willing to tolerate some friction, they can set up a client (whether branded as a backup app or just a regular messaging app) on a secondary backup device.

Definitely an option, although there are large parts of the world where most users have a single device. It feels like this would only be an option as a temporary means of scope-cutting, but wouldn't work as a complete long-term solution.

XIPs/xip-31-message-backup-request-system.md
richardhuaaa: > Why would they go through the trouble of using the backup service to exfiltrate messages without user consent when they can just use the SDK directly to get them?

To clarify, the goal is not to protect against malicious backup providers, just careless ones which may then serve backups to a malicious requester.

> We should figure out how to enforce it, but it's a sensible idea.

A possible idea is to include this as part of the signing step for a new installation. For example, signing step looks like this:

```
XMTP: Grant Messaging Access

Current Time: <current time and local timezone>
Installation Key: <Curve25519PublicKey:Ed25519PublicKey>

Permissions:
- Full message history
```

When a provider receives a backup request, they can look for this signature, verify that 'full message history' has been granted, and only serve backups if the local clock is within 10 mins of what the user signed for.

XIPs/xip-31-message-backup-request-system.md
```diff
@@ -0,0 +1,454 @@
...
+    RequestNotFound,
+    InvalidRequest,
+    NetworkError,
+    EncryptionError
+}
+```
+
+## Rationale
+
+There are many other potential solutions to message portability listed below. I am proposing this particular solution because it solves the key user problems stated above with the minimal amount of developer friction. Many of the alternatives require substantial cooperation from application developers, who may care more about shipping new features than supporting competing applications. Developers have strong incentives to receive backups, but limited incentives to provide them.
```
trebor-yatska, line 364: Any potential network fees would be the responsibility of the Remote Message Backup provider as the sender?

PR #32, opened 2023-11-08
Update Idea phase with more details

jhaaaa:


XIPs/xip-0-purpose-process.md
```diff
@@ -38,13 +38,21 @@ An XIP must meet certain minimum criteria. It must be a clear and complete descr
...
+**To get started with your idea**:
 
-To get started, you can **submit your idea as a new topic in the [XIP Ideas category in XMTP's GitHub Discussions](https://github.com/orgs/xmtp/discussions/categories/xip-ideas)**.
+1. Submit your idea as a new topic in the [Ideas & Improvements category](https://community.xmtp.org/c/development/ideas/54) in the XMTP Community Forums.
+
+1. If you've built a prototype for your idea, provide access to it in your topic. A prototype can help de-risk your idea and give people something to anchor discussions around.
+
+1. Be sure to include the date by which you want people to provide feedback. A good standard is giving people ~2 weeks, including 2 weekends.
+
+1. Share your idea with the community to get feedback. For example, beyond posting to the XMTP Community Forums, you might want to share a link to your idea topic with your networks on other platforms.
```
jhaaaa, line 53: I am curious about how XMTP Editors and XMTP Core Developers can help the idea author amplify their idea, if they want help. For example, with the idea author's permission, an XMTP Core Developer with a Twitter following, might share it to get more people interested? Is that helpful or appropriate?

XIPs/xip-0-purpose-process.md
```diff
@@ -58,8 +66,8 @@ After vetting the idea as described above, your next responsibility is to presen
 
 ### XIP Process
 
-- **Idea** - An idea that is pre-draft. This is not tracked within the XIP Repository.
-- **Draft** - The first formally tracked stage of an XIP in development. An XIP is merged by an XIP Editor into the XIP repository when properly formatted.
+- **Idea** - An idea that is pre-draft. This is not tracked within the XIP Repository. To learn more, see [Get started with an idea](#get-started-with-an-idea).
+- **Draft** - The first formally tracked stage of an XIP in development. The idea author (now the XIP Author) creates the XIP Draft based on the guidelines and template discussed in the following sections. The XIP Draft is merged by an XIP Editor into the XIP repository when properly formatted.
```
jhaaaa, line 70: FYI - Draft to Review - will be the next phase we refine, as we did with Idea to Draft.

XIPs/xip-0-purpose-process.md
```diff
@@ -38,13 +38,21 @@ An XIP must meet certain minimum criteria. It must be a clear and complete descr
...
-To get started, you can **submit your idea as a new topic in the [XIP Ideas category in XMTP's GitHub Discussions](https://github.com/orgs/xmtp/discussions/categories/xip-ideas)**.
+1. Submit your idea as a new topic in the [Ideas & Improvements category](https://community.xmtp.org/c/development/ideas/54) in the XMTP Community Forums.
+
+1. If you've built a prototype for your idea, provide access to it in your topic. A prototype can help de-risk your idea and give people something to anchor discussions around.
+
+1. Be sure to include the date by which you want people to provide feedback. A good standard is giving people ~2 weeks, including 2 weekends.
+
+1. Share your idea with the community to get feedback. For example, beyond posting to the XMTP Community Forums, you might want to share a link to your idea topic with your networks on other platforms.
+
+1. After you feel your idea has received enough feedback, you can create an XIP Draft for it, as described in the next section. A good standard for "enough feedback" is at least one comment from an XMTP community member and one from an *XMTP Core Developer* (currently the XMTP Labs team).
```
daria-github, line 55: Should this be phrased as something more along the lines of "After feedback date has passed and there is at least 1 internal and 1 external review" -- just thinking we may want to make this step more concrete

XIPs/xip-0-purpose-process.md
daria-github: I think so! More feedback the better for this stage.

daria-github, line 53: should we give example platform links here?

jhaaaa: After processing for a lil bit - I think the amplification action is permissionless - but whether employees of XMTP Labs should amplify is a question of credible neutrality that we'll need to discuss. Also, if an idea is amplified using the XMTP protocol X.com account - I believe it must amplify all or none to be credibly neutral. More to dig into here. But I think the doc as-is works.

XIPs/xip-0-purpose-process.md
jhaaaa: Because this documentation is meant for the general public - I was thinking that they would not understand what we mean by 1 internal and 1 external review. <3 For the general public, I believe internal translates to "XMTP Core Developers" - who are designated as members of the XMTP Labs team in Discourse, and external translates to an XMTP community member. Does this make sense? =)

XIPs/xip-0-purpose-process.md
daria-github: Makes total sense!

XIPs/xip-0-purpose-process.md
daria-github: Yes! Absolutely. Those are much better choices given the audience - thanks for the clarification!

daria-github (approved):

XIPs/xip-0-purpose-process.md
jhaaaa: I think it is OK to leave this open to authors? They will know which platforms are best to reach their networks.

PR #53, opened 2024-04-10
Add XIP-46: Multi-Wallet Identity

jhaaaa:


XIPs/xip-46-multi-wallet-identity.md
```diff
@@ -0,0 +1,520 @@
...
+  // The actual signature bytes
+  bytes signature = 3;
+}
+
+// An existing address on xmtpv2 may have already signed a legacy identity key
+// of type SignedPublicKey via the 'Create Identity' signature.
+// For migration to xmtpv3, the legacy key is permitted to sign on behalf of the
+// address to create a matching xmtpv3 installation key.
+// This signature type can ONLY be used for CreateInbox and AddAssociation
+// payloads, and can only be used once in xmtpv3.
```
scotboyd, line 246: Does "only once" mean you have to concatenate the CreateInbox and AddAssociation into a single IdentityUpdate?

XIPs/xip-46-multi-wallet-identity.md
```diff
@@ -0,0 +1,520 @@
...
+```
+
+The rules for processing this signature are as follows:
+
+1. The recovered address of the ‘Create Identity’ signature in the `delegated_key` is the address that this action is being performed on behalf of.
+2. The ‘Create Identity’ signature of the `delegated_key` is implicitly treated as a `CreateInbox` action, with a nonce of 0.
+   1. During server-side validation, if the address is already associated with an Inbox ID (has pre-existing [Address Log](https://www.notion.so/Multi-Wallet-Identity-XIP-ba30e6ddb30c45acb2dc4ff9eccae7ee?pvs=21) entries), the whole `IdentityUpdate` is invalid.
+3. The `delegated_key` may be used to perform a single `AddAssociation` action for a new installation key.
+
+This delegated signing action may only be performed in one `InboxUpdate`. This is enforced via the `seen_signatures` HashSet described in previous sections.
```
scotboyd, line 520: Do you mean IdentityUpdate?

XIPs/xip-46-multi-wallet-identity.md
```diff
@@ -0,0 +1,520 @@
...
+  }
+}
+```
+
+The user may perform identity updates, such as adding or removing wallets and installations to an Inbox ID, which are signed by these keys. When a member is associated with an Inbox ID, it is assigned a role. The permissions granted to each role are defined as follows.
+
+| Permission                                                  | Associated address | Installation key | Recovery address |
+| ----------------------------------------------------------- | ------------------ | ---------------- | ---------------- |
+| Is typically a wallet                                       | Yes                | No               | Yes              |
+| Addressable for incoming messages                           | Yes                | No               | No               |
```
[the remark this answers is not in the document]
neekolas: Maybe it could be clearer, but we're saying that the installation key itself isn't addressable. 

An inbox_id with an installation key is.

PR #65, opened 2024-06-28
Add XIP-31: Message History request system

jhaaaa:


XIPs/xip-31-message-history.md
```diff
@@ -0,0 +1,435 @@
...
+   5e. Delete the `MessageHistoryBackupResponse` and all associated credentials.
+
+   5f. Set the status of the `MessageHistoryBackupRequest` to `Applied`.
+
+#### Flow for Message Backup Providers
+
+1. Receive a `MessageHistoryBackupRequest` as part of the normal message receiving flow.
+1. Retrieve the contact for the installation that sent the message and ensure it has not been revoked. If revoked, ignore.
+1. Ensure the installation that sent the message has a contact signed by the same blockchain account as the current user. If not, ignore.
+1. Convert all messages in the local database into a Message Backup File (maybe we want to chunk here?)
```
jhaaaa, line 102: OK to remove the question? Has this been resolved?

XIPs/xip-31-message-history.md
```diff
@@ -0,0 +1,435 @@
...
+- An app that allows users to talk to customer support needs access to only the newly created conversation with the support agent.
+- A marketplace app that wants to allow buyers and sellers to communicate about a particular item. These would only be new conversations and may not care about preexisting messages.
+
+### Design goals
+
+1. Message history synchronization should not compromise the Forward Secrecy or Post Compromise Security properties of XMTP V3. We expect the majority of consumer apps using XMTP to have some form of message history syncing enabled.
+1. Message history synchronization should be opt-in for developers.
+1. It should be simple for developers to implement message history synchronization correctly. Implementation work should be focused on app-specific UX choices, with the SDK responsible for core transport and security decisions.
+1. Compatibility between apps is expected. Apps should be able to share history irrespective of operating system, SDK version, or device type (web vs. mobile vs. server).
+1. Even if developers of some popular apps choose to not support message history synchronization, it should be simple for users to opt enable themselves.
```
jhaaaa, line 40: Correct edit?

```suggestion
1. Even if developers of some popular apps choose to not support message history synchronization, it should be simple for users to opt-in to enable it for themselves.
```

XIPs/xip-31-message-history.md
```diff
@@ -0,0 +1,435 @@
...
+```
+
+`GET /backups/$DOWNLOAD_ID`:
+Returns the uploaded file matching the ID
+
+It would be the responsibility of the Backup Storage Provider to authenticate requests to `/backups` and mitigate abuse. Uploaded files would only need to be stored for maybe 72 hours before they could be safely purged, as backups are meant to be temporary storage. We could also just delete the file after it had been downloaded once.
+
+XMTP Labs would provide a reference implementation of a Backup Storage Provider.
+
+I am also proposing that XMTP Labs run a Backup Storage Provider as a free public good for the next two years, at which point this functionality would become part of an ecosystem of third-party gateway service providers.
```
jhaaaa, line 167: OK to leave this in?

XIPs/xip-31-message-history.md
```diff
@@ -0,0 +1,435 @@
...
+##### XMTP Network Advantages
+
+- No new infrastructure required. Everything stays on the XMTP network
+
+##### XMTP Network Disadvantages
+
+- The XMTP network was not designed for storage of large files. The network currently has a 1MB per message limit so backups would be divided across potentially hundreds of chunks.
+- Costly as the network decentralizes and fees are added to the network. Given that alternative storage providers can be used in the current proposal, the benefits to decentralization are limited.
+
+I could get on board with this proposal if others felt strongly that we shouldn't create new infrastructure. While backups are an odd fit for the XMTP transport network, it would work in the short term and be easier to bootstrap.
```
jhaaaa, line 360: OK to leave this in? Do we want to reword?

XIPs/xip-31-message-history.md
rygine: yes, OK to remove the question

XIPs/xip-31-message-history.md
rygine: yes, we are already running this service

XIPs/xip-31-message-history.md
rygine: yes, i think it can be removed. i'll do that in a commit.

XIPs/xip-31-message-history.md
```diff
@@ -0,0 +1,411 @@
...
+1. Reply to the message with a `MessageHistoryBackupResponse` containing the encryption details and the `backupUrl` provided by the Backup Storage Provider.
+1. Delete the `MessageHistoryBackupRequest`, the local database dump, and all the encryption keys.
+
+#### End-to-end flow
+
+![Sequence diagram illustrating the end-to-end flow for message backup providers|690x410](./assets/xip-31/message-backup-providers-flow.png)
+
+### Backup Account Files
+
+Backup account files are designed as a solution for users who have lost access to all of their XMTP apps. To create a Backup Account File, an app will generate a new set of XMTP keys and register the installation on the XMTP network. The installation contact published to the network will be set with a type of `backup`, which instructs all well-behaved clients to disregard any messages from this installation other than `MessageHistoryBackupResponse`s. This means that a compromised backup installation cannot be used to send messages on the user's behalf, which makes these files a less valuable target for attack.
```
rygine, line 112: i didn't see any mention of a `backup` contact in the code. is this something we want to pursue? @neekolas

XIPs/xip-31-message-history.md
```diff
@@ -0,0 +1,411 @@
...
+```json
+{
+  "downloadUrl": "https://backupproviderdomain.com/files/some-long-unguessable-download-id"
+}
+```
+
+`GET /files/$DOWNLOAD_ID`:
+Returns the uploaded file matching the ID
+
+It would be the responsibility of the Backup Storage Provider to authenticate requests to `/upload` and mitigate abuse. Uploaded files would only need to be stored for maybe 72 hours before they could be safely purged, as backups are meant to be temporary storage. We could also just delete the file after it had been downloaded once.
```
rygine, line 161: @neekolas should we implement this in the reference server?

XIPs/xip-31-message-history.md
neekolas: It's something we're shelving for now. Nothing is preventing us from adding the backup contact concept later. From the SDK perspective a "backup installation" is just another installation and the history request system would work the same. All the complexity lives at the app level.

XIPs/xip-31-message-history.md
neekolas: Not urgent, but yes it's a pending TODO item for that server.

Ideally we move the backups off the filesystem altogether, so I wouldn't worry about expiring them before we do that.

PR #73, opened 2024-11-27
Update xip-46-multi-wallet-identity.md

neekolas:
Clarify role of recovery address, and permissions of installation keys


jhaaaa (approved):
Thank you so much for this, @neekolas - so very helpful! LGTM!

PR #74, opened 2024-12-03
Queryable Content Types in Rust

cameronvoell:
Adding initial draft

To do:
- [X] Finish Backwards Compatibility
- [X] Test cases
- [X] Add links for Reference Implementation
- [X] Security Consideration
- [x] Address lint warnings

### DIFF
--- a/XIPs/xip-48-content-type-local-db-integration.md
+++ b/XIPs/xip-48-content-type-local-db-integration.md
@@ -16,7 +16,11 @@ This XIP proposes a way of managing XMTP content types in protobufs and rust in 
 
 ## Motivation
 
-During the upgrade from XMTP V2 (Direct Messaging Only) to XMTP V3 (Groups via the MLS protocol), one developer experience improvement was that the new "V3" SDKs manage a local SQLITE database containing all of a user's groups and messages. At the same time, our core "libxmtp" library that manages all the database logic remained completely agnostic to the message "content types" that were being stored in the database. While this separation of content types from the database logic was great for keeping complexity down during initial developement, it has made some quieries that would be common in a production consumer messaging app impossible without forcing developers to implement their own extra local persistent data management. An example of this is when the SDK returnsall message types in order received for a group, how can you efficiently render the emoji reactions, replies and read reciepts for each message in your UI? In order to address these types of queries, this XIP proposes a way to integrate XMTP content types with our core library and local SQLITE storage.
+During the upgrade from XMTP V2 (Direct Messaging Only) to XMTP V3 (Groups via the MLS protocol), one developer experience improvement was that the new "V3" SDKs manage a local SQLITE database containing all of a user's groups and messages. At the same time, our core "libxmtp" library that manages all the database logic remained completely agnostic to the message "content types" that were being stored in the database. The fact that our local database code is content type agnostic has made some queries that would be common in a production consumer messaging app impossible without forcing developers to implement their own extra local persistent data management. This XIP proposes a way to integrate XMTP content types with our core library and local SQLITE storage. This solution will enable the following new SDK functions:
+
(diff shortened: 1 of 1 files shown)


XIPs/xip-48-content-type-local-db-integration.md
```diff
@@ -0,0 +1,338 @@
...
+
+Users on older SDK versions both using JSON encoding will be fine, and new users both using protobuf encoding will be fine. The cases where backward compatibility will be an issue are the following:
+
+#### 1. An older XMTP SDK version receives a message with a new protobuf content type
+
+In this case, your SDK version should recognize that it does not know how to decode the content type because of it's higher ContentTypeId version, and it will instead use the fallback string of the EncodedContent wrapper struct.
+
+#### 2. An older XMTP SDK version sends messages with old JSON content types to users on newer SDK versions
+
+One option is for integrator targeted SDKs to retain code for decoding older JSON content types. A second option is to update integrator targeted SDKs to only support the latest protobuf based content types that will work in new content type based message functions, and to only use fallback text for messages that are not compatible with the latest content types.
```
neekolas, line 282: Agree that backwards compatibility isn't the top priority. The only thing that makes me lean towards the first option is how straightforward it should be.

Will make integrators lives easier if they can trust that any reaction they see will be decoded properly. I worry that if we don't integrators will see the fallback from older messages and try to parse it or manually decode the content.

cameronvoell: > I worry that if we don't integrators will see the fallback from older messages and try to parse it or manually decode the content.

Good point, and I agree it shouldnt be too difficult. Updated here https://github.com/xmtp/XIPs/pull/74/commits/f403869d734a4ff58f1379210a79df34d29d079c. (also simplified the backwards compatability section since it can be summarized as two cases instead of four)

jhaaaa (approved):
Thank you so much for this XIP, @cameronvoell! Made just a few XIP copyedits and posted to https://community.xmtp.org/t/xip-48-queryable-content-types-in-rust/845.

One note for a future iteration is whether you'd like to add a "threat model" section. So sorry I didn't mention it earlier! I need to update the XIP template to include this section.

XIPs/xip-48-content-type-local-db-integration.md
```diff
@@ -0,0 +1,326 @@
...
+
+\* In the case of content types managed by the [XMTP GitHub organization](https://github.com/xmtp) the authority would be contributors/admins of the relevant GitHub repos.
+
+### Threat model
+
+As usual, users and developers should aspire to be using the latest XMTP SDK versions whenever possible, but you can never be sure whether inboxes you are chatting with on XMTP are using older or modified versions of the SDK. This means that users could configure their messages to specify that they are sending a certain content type, but the format, contents, or intentions of the actual data in the message could be something different. Fortunately, we haven't identified any security issues that would arise from this possibility, but it is something developers utilizing "complex decoding or presentation logic" as mentioned in the quote above, should be aware of. If any more specific threats or example of misuse arise, this section will be updated.
```
jhaaaa, line 322:
```suggestion
As usual, users and developers should aspire to use the latest XMTP SDK versions whenever possible, but you can never be sure whether inboxes you are chatting with on XMTP are using older or modified versions of the SDK. This means that users could configure their messages to specify that they are sending a certain content type, but the format, contents, or intentions of the actual data in the message could be something different. Fortunately, we haven't identified any security issues that would arise from this possibility, but it is something developers utilizing "complex decoding or presentation logic" as mentioned in the quote above, should be aware of. If any more specific threats or examples of misuse arise, this section will be updated.
```

jhaaaa (approved):
Thank you next-level @cameronvoell ! Added just a couple of suggested edits. 🫡

PR #84, opened 2025-01-29
XIP-53: XMTP V2 deprecation plan

jhaaaa:


XIPs/xip-53-v2-deprecation.md
```diff
@@ -0,0 +1,76 @@
...
+author: Naomi Plasterer (@nplasterer)
+discussions-to: TBD
+status: Draft
+type: Standards
+category: XRC
+created: 2025-01-29
+---
+
+## Abstract
+
```
jhaaaa, line 14: Took feedback @cameronvoell in the Notion and rewrote a bit - WDYT?

XIPs/xip-53-v2-deprecation.md
```diff
@@ -0,0 +1,76 @@
...
+XMTP V3 provides the framework for a more robust messaging ecosystem—supporting stronger security guarantees with MLS, modern features like group chats, and decentralization. These enhancements weren’t efficiently and effectively feasible with XMTP V2, which also can’t migrate seamlessly to V3.
+
+To encourage adoption of XMTP V3 and ensure the network remains secure and future-proof, we provide this plan to remove the ability to write to XMTP V2. This step discourages reliance on outdated infrastructure and fosters a unified community transition to the more advanced V3 network.
+
+## Motivation
+
```
jhaaaa, line 20: Also took great feedback from @cameronvoell here as well - and rewrote a bit. WDYT?

XIPs/xip-53-v2-deprecation.md
```diff
@@ -0,0 +1,80 @@
...
+Now that XMTP V3 is fully featured and developer-ready, it’s the ideal time to move V2 interactions to a more future-proof environment that provides a sustainable, long-term solution for the XMTP community. By transitioning to V3, these interactions can remain stable and continue to compound into the future.
+
+## Specification
+
+**January 1, 2025**: Developers should start upgrading to XMTP V3, moving off any V2-compatible versions of XMTP SDKs. We recommend always updating your app to use the latest version of the SDK to ensure that you'll be on a stable release when it's ready.
+
```
jhaaaa, line 26: Did a bit of research on brownouts and provided the following proposal. 

- They recommended using consistent days for ease of recall - so perhaps every Tues and Thur, or some other consistent set of days of the week.

- They shared a commonly used brownout time window of 02:00 AM - 06:00 AM UTC - so this reflects that.

- Communication is key and we will communicate repeatedly before the brownout! Do we expect to provide real-time support to any devs who haven't received the memo and are wondering what happened? If so, we should have a planned and equitable schedule for Ephemera engineers to handle support in real-time during off hours. Or perhaps there is a document we can post to help these devs in lieu of live support?

- Curious about what these brownout times are used for -- beyond motivating devs to move to V3. Is this brownout providing some kind of testing period as well? For example, will an app dev use a brownout period to test how their app performs without V2 writes? Perhaps there is some doc we can write about how to test this without waiting for the brownout?

XIPs/xip-53-v2-deprecation.md
```diff
@@ -0,0 +1,80 @@
...
+- 09:00 AM - 01:00 PM ICT (next day)
+
+East Asia
+- 10:00 AM - 02:00 PM CST (next day)
+- 11:00 AM - 03:00 PM JST (next day)
+
+UTC
+- 02:00 AM - 06:00 AM UTC (next day) 
+
+**May 1, 2025** - At 07:00 PM PDT, XMTP V2 will be permanently set to read-only mode. No more writes to the V2 network will be allowed.
```
jhaaaa, line 62: Added a specific time the brownout will start. I set it to 7 pm PDT b/c it is consistent with the start of brownout time windows above.

XIPs/xip-53-v2-deprecation.md
```diff
@@ -0,0 +1,80 @@
...
+
+However, users can still access their V2 messages in a read-only capacity using [legacy.xmtp.chat](https://legacy.xmtp.chat/).
+
+## Reference implementation
+
+[Upgrade to XMTP V3](https://docs.xmtp.org/upgrade-to-v3)
+
+## Security considerations
+
+Identities on XMTP V3 are more secure than on XMTP V2. Upgrade your app to use XMTP V3 to give your users much stronger security guarantees.
```
jhaaaa, line 76: Naomi had a great suggestion that we link to a doc with the XMTP v3 security guarantees. We have been linking to a forum post for this info - but it should have a doc so I am going to create that and we can link to it.

XIPs/xip-53-v2-deprecation.md
nplasterer: The motivation behind these brownouts is to alert developers who may not know they are using V2 or are unaware of our decision to deprecate V2 because no prior communication reached them. It's a hard flag to alert them. 

Examples of this are
- Developer is on a version of the SDK that has a enableV3 flag set to true. While yes this enabled a hybrid version of the sdk it did not move them on to V3 entirely and there experience will break when we deprecate V2.
- Developer who published a bot on an old version of the SDK that was a hybrid and forgot about it. This bot will break because it's not on a pure V3 version.
- Developer who was unable to be reached by any of our comms and discovers it from users of the app saying the experience is broken and then goes to our website to file an issue and then we can educate.

I will also note we might be able to set the error message to be very clear that this is a brownout and link to documentation.

nplasterer (approved):

XIPs/xip-53-v2-deprecation.md
cameronvoell: looks great 🙏

cameronvoell (approved):

PR #85, opened 2025-01-30
Pull originator ID from AAD in XIP-49

mkysel:
This PR removes the responsibility of the client to determine a node to send the message to.

Instead, it becomes the payer's decision how to route messages in the system.

Remove the load balancing code and recommend stable hashing (ring hashing) instead.


mchenani (approved):

jhaaaa (approved):
thank you so much @mkysel !

fbac (approved):

PR #88, opened 2025-02-02
Update xip-52-group-announcement.md

nplasterer:
Update to make it clearer that it is not a breaking change

### DIFF
--- a/XIPs/xip-52-group-announcement.md
+++ b/XIPs/xip-52-group-announcement.md
@@ -33,7 +33,7 @@ This would be the similar to a description of a group, but each app can display 
 
 - Any group that existed before the `groupAnnouncement` feature was added won't have an entry for `groupAnnouncement` in its mutable metadata. There will also be no predefined permission policy for who is allowed to update the `groupAnnouncement` metadata field.
 - If a user on a newer version of the protocol calls `updateGroupAnnouncement` on a pre-existing group, they're trying to set a value for `groupAnnouncement`, even though no permission policy exists for it yet. In this case, default permission policies will apply to determine whether the update is valid.
-If someone on an old version is invited to a group from someone on the new version, and so there is a permission policy set for groupAnnouncement.  Users on the older versions will actually be able to read the permission policy because it is fetched using that string key.
+- If a user on an older version of the protocol is invited to a group created by a user on the newer version, then the group will have a permission policy set for `groupAnnouncement`. Even though the older version of the protocol doesn't support setting announcements, it can still read the permission policy because it can be fetched using that string key.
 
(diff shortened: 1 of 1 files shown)


XIPs/xip-52-group-announcement.md
```diff
@@ -31,7 +31,9 @@ This would be the similar to a description of a group, but each app can display
 
 ## Backward compatibility
 
-This is a breaking change. To use this new `announcement` metadata field, all group members must be using an app that uses an SDK version that supports the metadata.
+If a group exists before we add group announcements. then it will have no value for group announcement in mutable metadata and no permission policy for who is allowed to update a mutable metadata field with the "groupAnnouncement" key.
```
jhaaaa, line 34: How about something like:

```suggestion
- Any group that existed before the `groupAnnouncement` feature was added won't have an entry for `groupAnnouncement` in its mutable metadata. There will also be no predefined permission policy for who is allowed to update the `groupAnnouncement` metadata field.
```

XIPs/xip-52-group-announcement.md
```diff
@@ -31,7 +31,9 @@ This would be the similar to a description of a group, but each app can display
...
+Then someone on the new version calls updateGroupAnnouncement on that group to set that value, even though there is no existing permission policy in that group because it is already created. In that case we have default permission policies to evaluate if the metadata update is valid or not.
```
jhaaaa, line 35: How about something like:

```suggestion
- If a user on a newer version of the protocol calls `updateGroupAnnouncement` on a pre-existing group, they're trying to set a value for `groupAnnouncement`, even though no permission policy exists for it yet. In this case, default permission policies will apply to determine whether the update is valid.
```

XIPs/xip-52-group-announcement.md
```diff
@@ -31,7 +31,9 @@ This would be the similar to a description of a group, but each app can display
...
+If someone on an old version is invited to a group from someone on the new version, and so there is a permission policy set for groupAnnouncement.  Users on the older versions will actually be able to read the permission policy because it is fetched using that string key.
```
jhaaaa, line 36: How about something like:

```suggestion
- If a user on an older version of the protocol is invited to a group created by a user on the newer version, then the group will have a permission policy set for `groupAnnouncement`. Even though the older version of the protocol doesn't support setting announcements, it can still read the permission policy because it's stored as a key-value pair in the metadata.
```

jhaaaa:
Thank you for this @nplasterer! Made just a few tech writer 🤓 edits... WDYT?

nplasterer:
Looks good! thanks for the edits. 👍

PR #110, opened 2025-05-29
Initial revision for this XIP, have a few questions to resolve before …

cboscolo:
…final

### DIFF
--- a/XIPs/xip-66-atproto-identity.md
+++ b/XIPs/xip-66-atproto-identity.md
@@ -1,0 +1,126 @@
+---
+xip: 66
+title: AT Protocol Identity
+description: Proposes a way to add an AT Protocol DID as an MemberIdentifir for an InboxID
+authors: Chris Boscolo (@ccbscolo)
+status: Draft
+type: Standards
+category: XRC
+created: 2025-05-25
+---
+
+## Abstract
+
+[XIP-66](https://github.com/xmtp/XIPs/blob/main/XIPs/xip-66-atproto-identity.md) defines a way to add an AT Protocol identity as an associated identity for an XMTP Inbox.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


XIPs/xip-66-atproto-identity.md
```diff
@@ -0,0 +1,126 @@
...
+
+There are two challenges with AT Protocol identities the make it different than either a Passkey or an Ethereum Wallet.
+
+1. An ATProto DID (account) is an [W3C DID](https://atproto.com/specs/did) (eg. `did:plc:bq5ygkxmmbae7n6z34neuaih`). The DID itself may not be self-certifying the way a wallet EOA or public key is. Instead, the DID Document contains the public key designated by the user to authenticate writes to the [Personal Data Store (PDS)](https://atproto.com/guides/data-repos). It's assumed that there is secure way to query the DID Document for the DID. In order to prove control over the AT Protocol identity, client software needs to lookup the DID Document associated with the ATProto DID and read the `#atproto` public key from the `verificationMethod` element, then use this  key to verify signatures of data commited to user's PDS. Similar to Smart Contract Wallet address, verification of signatures cannot be performed offline. One ramification of this, is that an ATProto DID would not be suitable as an `MemberIdentifier` that can add installations (key pairs) as children. An ATProto DID would only serve as an "also known ss" identifier for the XMTP Inbox ID.
```
richardhuaaa, line 28: This XIP is well-written, thank you! I couldn't find this in the ATProto docs, but is this public key mutable?

- If immutable, there may be some optimizations such as caching that could also be explored.
- If mutable, it might be helpful to flesh out some details of how older signatures are handled when the public key changes. We had to work through a similar issue with smart contract signatures.

I'm assuming it's the latter, seeing as it's hosted on a webserver?

